PR #119, opened 2024-05-30
Add Ray version for multi file process

xuechendi:
## Description

Add ray version of upload file and ingestion to redis

## Issues

## Type of change

## Dependencies

Ray, pyarrow, virtualenv

## Tests

tested with uploading and ingesting 1000 pdf files


xuechendi:
1. a test file added, took 1000 secs to complete 1000 pdf ingestion (around 500MB)
image

2. status will be saved as csv files
image

image

comps/dataprep/redis/prepare_doc_redis_on_ray.py
```diff
@@ -0,0 +1,225 @@
...
+from ray.data.block import Block
+from ray.data.datasource import FileBasedDatasource
+
+from comps import DocPath, opea_microservices, register_microservice
+from comps.dataprep.utils import document_loader, timeout
+
+tei_embedding_endpoint = os.getenv("TEI_ENDPOINT")
+
+
+def prepare_env(enable_ray=False, pip_requirements=None):
```
ftian1, line 47: ray initialization should be moved after line 224

xuechendi:
@ftian1 @lvliang-intel, please help to review this PR.
Here is the latest performance

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/chendixu/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/chendixu/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D" lang=en-US>


  | Original | w/ ray - now
-- | -- | --
(secs) | (secs)
process 1000 PDF     (558MB) | 5919.54 | 1281.981
Process   10 PDF | 41.617 | 34.345
process single file | 4.32 | 17.736
  |   |  
url from a news website | Original | w/ ray
(secs) | (secs)
process 1000 link | 3369.78 | 263   secs
Process 10 links | 42   secs | 36   secs
process single link | 3.67   secs | 30   secs



</body>

</html>

xuechendi:
My question is:
since the docker option is set "prepare_doc_redis.py" as entry point, I can
      option1:  commit the ray code to 'prepare_doc_redis.py' file?
      option 2: create a very similiar dockerfile only change the entry point to 'prepare_doc_redis_on_ray.py'
Would like to know your preference.

xuechendi:
@ftian1 , Based on our discussion, I create a new microservice for Ray data prep, test is added to tests/test_ingestion.sh

comps/dataprep/utils.py
```diff
@@ -163,6 +214,7 @@ def load_svg(svg_path):
     return text
 
 
+@timeout(600)
```
ftian1, line 217: @lvliang-intel  pls help to check if this part is ok for you?

@xuechendi my major concern here is even timeout reaches, there is no benefit for user as the current operation will only dump message in server side, user aren't aware of this.

comps/dataprep/redis/prepare_doc_redis_on_ray.py
```diff
@@ -0,0 +1,304 @@
...
+                enable_ray = True
+            prepare_env(enable_ray=enable_ray)
+            num_cpus = get_max_cpus(len(link_list))
+            print(f"per task num_cpus: {num_cpus}")
+            ingest_link_to_redis(link_list, enable_ray=enable_ray, num_cpus=num_cpus)
+            return {"status": 200, "message": "Data preparation succeeded"}
+        except json.JSONDecodeError:
+            raise HTTPException(status_code=400, detail="Invalid JSON format for link_list.")
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=f"An error occurred: {e}")
```
xuechendi, line 299: @ftian1 , all error including timeout will be returned through 400 for python side. For Ray side, since this is for large data processing, I can return which files didn't been correctly processed in success "200"

comps/dataprep/utils.py
xuechendi: @ftian1, 
1. for file size less than 10, it will go python mode, and any failure including timeout will be return in error 400 description. 
2. For any file size more than 10, it will go ray mode, yes, only server side captures the info of which files are failed. Since if we are handling 1000 or more files, returning such info to user may not be necessary.
(But if you would like to add that, I can make it return 200 but provide filename which is not successfully handled in description.If you think this works for you, I'll add this feature.)

ftian1: I just think the purpose of introducing timeout is used to capture those hang operation, right? shall it be deal with by k8s or by self? I am worried that there always has a case in which the file size is huge and request longer time than your hardcoded timeout value in code.

xuechendi: @ftian1, yes, most of distributed framework(ray, spark, etc) will provide a timeout mechanism for hanging tasks or tasks longer than expectation. 10min is a reasonable setting, if a single file/url processing takes more than 10 min, I think we should ask user to split or we should provide special treatment for it?
And for ray, if the timeout is caused by bad resource allocation, we can rerun the task based on the log.

I would suggested if we can test with actual enterprise usecase, we can make timeout configurable and also provide failover resume based on the user scenario.

xuechendi: @ftian1 , I added status return and provide timeout as a configurable environment.
* status will return to client
image

* Timeout setting
image

ftian1 (approved):

xuechendi:
@lvliang-intel , may you take a look of this PR?

lvliang-intel (approved):

PR #125, opened 2024-05-31
Enable microservice with ragas

XuhuiRen:
Enable evaluation for the [query, doc, answer, groundtruth] pairs with RAGAS based on the local langchain llm


comps/ragas/tgi/Dockerfile
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2024 Intel Corporation
```
chensuyue, line 1: Remove the old license head, the bot will help to apply the new one.

PR #181, opened 2024-06-14
Add LVMs LLaVA component

Spycsh:
## Description

Add LVMs LLaVA component

## Issues

n/a

## Type of change

## Dependencies

None
## Tests

tests/test_lvm.sh


lvliang-intel (approved):

comps/cores/mega/constants.py
```diff
@@ -26,6 +26,7 @@ class ServiceType(Enum):
     DATAPREP = 9
     UNDEFINED = 10
     RAGAS = 11
+    Img2txt = 12
```
hshen14, line 29: uppercase?

hshen14 (commented):
Shall we call "Img2txt" to "LVM"?  Image to text sounds too specific.

LVM means large visual model.

comps/cores/mega/constants.py
Spycsh: fixed

Spycsh:
> Shall we call "Img2txt" to "LVM"? Image to text sounds too specific.
> 
> LVM means large visual model.

Hi@hshen14 The original design here is to not only use LLaVA as the base model which convert image+prompt => text, but also adaptable to use image-captioning models such as [vit](https://huggingface.co/nlpconnect/vit-gpt2-image-captioning), [blip](https://huggingface.co/Salesforce/blip-image-captioning-large), which simply convert image=>text. So here I think img2txt is a more general term. What do you think?

chensuyue (approved):

XinyuYe-Intel (approved):

PR #205, opened 2024-06-18
Addition of a file change detection mechanism.

ZePan110:
## Description

## Issues

## Type of change

## Dependencies

## Tests


.github/workflows/test.yaml
```diff
@@ -0,0 +1,34 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for changed Dockerfile paths
+        run: |
+          merged_commit=$(git log -1 --format='%H')
```
mkbhanda, line 27: is hours granularity adequate?

chensuyue (approved):

.github/workflows/test.yaml
chensuyue: `-1`: This option limits the output to just the most recent commit.
`--format='%H'`: This specifies the format of the output. %H is a format specifier that outputs the full commit hash.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,55 @@
...
+                used_files+="$file "
+              fi
+            done
+          fi
+
+          if [[ "$is_use" == "TRUE" ]]; then
+            echo "Warning: Changed Dockerfile paths:"
+            echo "$used_files"
+            echo "Please modify the corresponding README in GenAIExamples repo and ask suyue.chen@intel.com for final confirmation."
+            exit 1
```
daisy-ycguo, line 54: I guess this job is a warning. Returning 1 will cause it failure. Maybe remove the error return statement and just return correctly.
What's more, if only the content of docker file is changed and not the file path, does the developer need to modify READMEs in GenAIExamples?
I think maybe the logic is like: find the removed (renamed) docker files in the changes log, and check if the original docker file path is used in READMEs in GenAIExamples.

daisy-ycguo (approved):
LGTM. Thank you.

kevinintel (approved):

PR #234, opened 2024-06-21
Fix bug between Retriever, TEI-embedding and VectorDB

zhlsunshine:
## Description

There are dependencies among Retriever, TEI-embedding and VectorDB,  the Retriever can get normal response if both EI-embedding and VectorDB are really up and running well.
However, the retriever would get error and this kind of error is always here if the EI-embedding or VectorDB are not available, and I encounter this situation in my gaudi node (Not the first time to deploy).  Please refer to log files in attach for more detail.

Fix solution:
I add the `check_embedding_endpoint` to make sure that the embedding endpoint is available, then begin to retrieve the data from VectorDB via TEI-embedding. And the problem is fixed, the Retriever log is shown as below:
image

Error log files:
[retriever-svc.log](https://github.com/user-attachments/files/15928527/retriever-svc.log)
[redis-vector-db.log](https://github.com/user-attachments/files/15928529/redis-vector-db.log)

## Issues

`n/a`

## Type of change

## Dependencies

`n/a`

## Tests


zhlsunshine:
We can not find this issue in Xeon, I think the reason is that Xeon is slower than Gaudi to handle the pdf file in retriever.

lvliang-intel (approved):

chensuyue (approved):

PR #235, opened 2024-06-23
[ASR] Fix asr microservice with Python run on SPR CPU

ctao456:
## Description

* Fixed Python path dependencies in `asr.py`.
* Added more instructions in `1. Start Microservice with Python (Option 1)` part of `comps/asr/readme.md` to make sure of proper run on Intel® SPR.
* Added expected output printouts in `comps/asr/readme.md`

## Issues

n/a

## Type of change

## Dependencies

n/a

## Tests

Tested on Intel® SPR CPU to make sure ASR microservice can be run properly.


lvliang-intel (approved):

comps/asr/asr.py
Spycsh, line 1: We do not need these two lines. Instead a build from source or explicitly export PYTHONPATH to `comps` is what we need to access the `comps`.

Spycsh (changes requested):
Edition to README looks good to me. Please remove the change to asr.py.

comps/asr/asr.py
ctao456: I tried build from source according to [Installation - Build from Source
](https://github.com/opea-project/GenAIComps), but the pip package name was `opea-comps` instead of `comps`. So `import comps` results in a `Module not found` error. 
image

Do you have inputs on this? Thanks.

Spycsh: Are you sure you have installed it correctly? I've tested it locally and there should not be a `Module not found` error. Please paste your error log.

Spycsh (approved):

PR #246, opened 2024-06-26
Fix and optimize the code for Gatway and Orchestrator

XuhuiRen:
## Description

Add more parameters ingestion in ChatQnA gatway, fix the code structure, insert the parameter into microservices.

## Issues

n/a

## Type of change

- New feature (non-breaking change which adds new functionality)
- Breaking change (fix or feature that would break existing design and interface)

### DIFF
--- a/comps/cores/mega/gateway.py
+++ b/comps/cores/mega/gateway.py
@@ -12,7 +12,7 @@ from ..proto.api_protocol import (
     ChatMessage,
     UsageInfo,
 )
-from ..proto.docarray import LLMParams, EmbedDoc768
+from ..proto.docarray import LLMParams, EmbedDoc768, SearchedDoc
 from .constants import MegaServiceEndpoint, ServiceRoleType, ServiceType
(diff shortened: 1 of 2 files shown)


comps/cores/mega/gateway.py
```diff
@@ -1,394 +1,411 @@
...
+class ChatQnAGateway(Gateway):
+    def __init__(self, megaservice, host="0.0.0.0", port=8888):
+        super().__init__(
+            megaservice, host, port, str(MegaServiceEndpoint.CHAT_QNA), ChatCompletionRequest, ChatCompletionResponse
+        )
+
+    async def handle_request(self, request: Request):
+        data = await request.json()
+        stream_opt = data.get("stream", True)
+        chat_request = ChatCompletionRequest.parse_obj(data)
```
lkk12014402, line 112: The `ChatCompletionRequest` doesn't define retriever and reranker parameters. How can we customize them?

PR #250, opened 2024-06-26
Fix unstable error in RAGAS

XuhuiRen:
## Description

Remove the unstable evaluation metric

## Issues

## Type of change


chensuyue:
@XuhuiRen do we still need this PR?

PR #266, opened 2024-06-30
optional vllm microservice container build

ashahba:
## Description

This PR adds optional `build-arg`'s for this Dockerfile: comps/llms/text-generation/vllm/docker/Dockerfile.microservice

## Issues

Current dockerfile always defaults to `gpu` Torch installation while `CPU` users might not necessarily need extra packages.
This leads to smaller containers for CPU users and faster build.

For users interested to do `gpu` builds all they need to do is build this image this way:
```
docker build --build-arg ARCH='gpu' -f comps/llms/text-generation/vllm/docker/Dockerfile.microservice . -t opea/llm-vllm:latest
```

## Type of change

- extra `build-arg` option `ARCH` to build `vllm microservice` container for either CPU or GPU.

## Dependencies

None

### DIFF
--- a/comps/llms/text-generation/vllm/docker/Dockerfile.microservice
+++ b/comps/llms/text-generation/vllm/docker/Dockerfile.microservice
@@ -1,8 +1,9 @@
-
 # Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 FROM langchain/langchain:latest
+
+ARG ARCH="cpu"
 
 RUN apt-get update -y && apt-get install -y --no-install-recommends --fix-missing \
     libgl1-mesa-glx \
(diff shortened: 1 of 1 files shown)


ftian1 (approved):

comps/llms/text-generation/vllm/docker/Dockerfile.microservice
```diff
@@ -1,9 +1,10 @@
-
 # Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 FROM langchain/langchain:latest
 
+ARG ARCH="cpu"  # Set this to "cpu" or "gpu"
```
zhlsunshine, line 6: If so, we need to maintain many versions of such Dockerfile, I suggest that there should be an environment variable to hold the value, and user can pass the specified ARCH value when building an image. Does it make sense?

XinyaoWa, line 6: Currently we don't support GPU, but it's OK to add this option for later extension.

XinyaoWa (approved):

comps/llms/text-generation/vllm/docker/Dockerfile.microservice
ashahba: I suggested this as a quick workaround, but doing this right needs to happen across board and for all Dockerfiles in all repositories.
That can be worked on but it's going to a RFC and needs some discussion first.

ashahba: I'd leave it.
As far as I know, some `pip` users don't like the `--extra-index-url` option and they can still build this with `gpu` if they want.

chensuyue (approved):

PR #274, opened 2024-07-04
Introduce 'entrypoint.sh' for some Containers

ashahba:
## Description

This PR introduces an `entrypoint.sh` script to be run at Container startup which is a very common practice by Container developers.

## Issues
There are times where more than a single command is required to be run at container entry point and having a script would simplify the Dockerfile design greatly.

## Type of change

- [N/A] Bug fix (non-breaking change which fixes an issue)
- [X] New feature - Pack multiple commands into an script and run that script at Container startup rather than a single command within the body of the container
- [N/A] Breaking change (fix or feature that would break existing design and interface)

## Dependencies

None.

## Tests

- Container builds are successful
- Running the `entrypoint.sh` script is relatively fast and does not introduce unacceptable wait time for prompt to be ready
- Running containers produces output similar to the following which is the expected result:
```
[2024-07-03 23:35:26,145] [    INFO] - CORS is enabled.
[2024-07-03 23:35:26,146] [    INFO] - Setting up HTTP server
[2024-07-03 23:35:26,146] [    INFO] - Uvicorn server setup on port 9000
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:9000 (Press CTRL+C to quit)
[2024-07-03 23:35:26,162] [    INFO] - HTTP server setup successful
```


comps/ragas/tgi/Dockerfile
```diff
@@ -1,26 +1,26 @@
...
-
-RUN pip install --no-cache-dir --upgrade pip && \
-    pip install --no-cache-dir -r /home/user/comps/ragas/tgi/requirements.txt
-
-ENV PYTHONPATH=$PYTHONPATH:/home/user
-
-WORKDIR /home/user/comps/ragas/tgi/
-
-ENTRYPOINT ["python", "llm.py"]
+FROM langchain/langchain:latest
```
ashahba, line 4: This `Dockerfile` had a lot of garbage characters and I had to run it through a linter and that's why it show all lines as green but only the last line is my actual change.

Spycsh (approved):

chensuyue (approved):

PR #279, opened 2024-07-06
Enlarge megaservice request timeout for supporting high concurrency

lvliang-intel:
## Description

Enlarge megaservice request timeout for supporting high concurrency

## Issues

`n/a`

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

## Dependencies

None

## Tests

Local test and pre-CI test.

### DIFF
--- a/comps/cores/mega/orchestrator.py
+++ b/comps/cores/mega/orchestrator.py
@@ -40,7 +40,8 @@ class ServiceOrchestrator(DAG):
     async def schedule(self, initial_inputs: Dict, llm_parameters: LLMParams = LLMParams()):
         result_dict = {}
 
-        async with aiohttp.ClientSession(trust_env=True) as session:
+        timeout = aiohttp.ClientTimeout(total=1000)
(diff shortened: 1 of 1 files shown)


comps/cores/mega/orchestrator.py
```diff
@@ -40,7 +40,8 @@ def flow_to(self, from_service, to_service):
     async def schedule(self, initial_inputs: Dict, llm_parameters: LLMParams = LLMParams()):
         result_dict = {}
 
-        async with aiohttp.ClientSession(trust_env=True) as session:
+        timeout = aiohttp.ClientTimeout(total=1000)
```
hshen14, line 43: is timeout 1000 still not enough? do we need to expose the parameter through env variable when the image is available?

lvliang-intel: 1000 seconds is sufficient even for high concurrency benchmark tests. I think we should keep this as internal information. It will confuse users if we export it as an environment variable.

letonghan (approved):

Spycsh (approved):

PR #280, opened 2024-07-08
Allow the Ollama microservice to be configurable with different models

lvliang-intel:
## Description

Allow the Ollama microservice to be configurable with different models

## Issues

`n/a`

## Type of change

## Dependencies

None

## Tests

Local test and pre-CI test.


hshen14 (approved):

comps/llms/text-generation/ollama/llm.py
```diff
@@ -21,7 +21,7 @@
 def llm_generate(input: LLMParamsDoc):
     ollama = Ollama(
```
lkk12014402, line 22: does the Ollama initialization have latency gap?

lvliang-intel: Good point, we will cover the test.

Spycsh: Yes, I verify and it sometimes has latency gap. Here is the default Ollama params https://github.com/ollama/ollama/blob/main/docs/modelfile.md

But I think it is safe to just use OpenAI default ones.

Spycsh: Also I guess this model name should be passed as input param. WIll adapt https://github.com/opea-project/GenAIComps/pull/309 later.

Spycsh (approved):

PR #313, opened 2024-07-16
Add local_embedding return 768 length to align with chatqna example

xuechendi:
## Description

Add a new local_embedding file which returns 768 embedding length instead of 1024 and use BAAI/bge-base-en-v1.5

## Issues

the local_embedding return and redis_retriever is not compatible at this moment.  

## Type of change

## Dependencies

No

## Tests

No


xuechendi:
@tianyil1 @lvliang-intel , may you take a look

lvliang-intel (approved):

kevinintel (approved):

PR #318, opened 2024-07-17
Add FAQGen gateway in core to support FAQGen Example, Adds FAQGen LLM service

yogeshmpandey:
## Description

Added components to support FAQGen Example. 

- FAQGen Gateway
- FAQGen LLM Endpoint

## Issues

n/a

## Dependencies

## Tests

Added tests to the tests folder.


amberjain1 (approved):

comps/cores/mega/gateway.py
```diff
@@ -400,3 +400,45 @@ async def handle_request(self, request: Request):
...
+        prompt = self._handle_message(chat_request.messages)
+        parameters = LLMParams(
+            max_new_tokens=chat_request.max_tokens if chat_request.max_tokens else 1024,
+            top_k=chat_request.top_k if chat_request.top_k else 10,
+            top_p=chat_request.top_p if chat_request.top_p else 0.95,
+            temperature=chat_request.temperature if chat_request.temperature else 0.01,
+            repetition_penalty=chat_request.presence_penalty if chat_request.presence_penalty else 1.03,
+            streaming=stream_opt,
+        )
+        result_dict = await self.megaservice.schedule(initial_inputs={"query": prompt}, llm_parameters=parameters)
```
Spycsh, line 424: Here the return values should be result_dict, runtime_graph

Please check https://github.com/opea-project/GenAIComps/blob/main/comps/cores/mega/gateway.py#L122C22-L122C36

comps/cores/mega/gateway.py
```diff
@@ -400,3 +400,45 @@ async def handle_request(self, request: Request):
...
+        for node, response in result_dict.items():
+            # Here it suppose the last microservice in the megaservice is LLM.
+            if (
+                isinstance(response, StreamingResponse)
+                and node == list(self.megaservice.services.keys())[-1]
+                and self.megaservice.services[node].service_type == ServiceType.LLM
+            ):
+                return response
+        last_node = self.megaservice.all_leaves()[-1]
```
Spycsh, line 433: Please replace `self.megaservice.all_leaves()[-1]` to `runtime_graph.all_leaves()[-1]`

Spycsh:
Hi @yogeshmpandey , thanks for contribution. This microservice works perfectly. Please check the comments and make sure the example also works with the latest code.

comps/cores/mega/gateway.py
yogeshmpandey: Fixed

comps/cores/mega/gateway.py
yogeshmpandey: Fixed

ftian1:
I am thinking if it's needed to introduce a new FAQGen type under comps/llm folder. the most of code is same with the one in comps/llm/summarization, just prompt_template is different, right? maybe a more appropriate way is to have a unique code with an extra template folder in which it contains many different templates to allow user to setup when starting the LLM docker.

yogeshmpandey:
> I am thinking if it's needed to introduce a new FAQGen type under comps/llm folder. the most of code is same with the one in comps/llm/summarization, just prompt_template is different, right? maybe a more appropriate way is to have a unique code with an extra template folder in which it contains many different templates to allow user to setup when starting the LLM docker.

Hi @ftian1 

I did not want to introduce any breaking change to other components hence added a new LLM endpoint,

I agree with this approach, the user can choose the specific prompt template defined by the environment variable. These templates can be read from a template file. 

Option #2
Another approach could be to allow users to provide a custom prompt. We can update the `ChatCompletionRequest` with an Optional `custom_prompt` field, enabling users to set their desired prompt. With this approach, the same backend can handle multiple tasks like summarization, FAQs, meeting summaries, etc. Additionally, this would provide an option for users to override the default custom prompt for summarization tasks.

ftian1 (approved):
merge it at first. we will unify the whole LLM instances together in the next release.

PR #342, opened 2024-07-24
add: Pathway vector store and retriever as LangChain component

berkecanrizai:
## Description

Adds Pathway Vector Store as a vector DB option.

Note that Pathway expects text documents for the retrieval, thus I added the embedder (both the local HF model and the `HuggingFaceHubEmbeddings(model=tei_embedding_endpoint)`) to the vector store initialization. 

Another additional note, Pathway connects to the input sources from the start, and does not allow `upsert` type operations via client. For now, I set the data source as the local `data` folder. It is possible to use any other data source, check out [here](https://pathway.com/developers/user-guide/connect/pathway-connectors).

## Type of change

Pathway vector DB and retriever.

## Dependencies

Pathway is needed if server is being started outside of the Docker.
Retriever and the other services do not need any change in the requirements.


comps/retrievers/langchain/pathway/retriever_pathway.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (C) 2024 Intel Corporation
```
hshen14, line 1: @chensuyue please check the license

comps/retrievers/langchain/pathway/Dockerfile.pathway
hshen14, line 1: @ashahba  please check the docker file

ftian1, line 1: suggest to move the file name from Dockerfile.pathway to Dockerfile as the folder name has included pathway

berkecanrizai: Hey, I renamed the file.

Spycsh (approved):

comps/retrievers/langchain/pathway/data/nke-10k-2023.pdf
lvliang-intel, line 1: Now the retrieval microservice doesn't need to ingest data during startup. Users can ingest data using the dataprep microservice. So we can remove the pdf file and let user do it in README.

comps/retrievers/langchain/pathway/pathway_vs.py
Spycsh, line 1: If I understand correctly, you place the vectorstore build/run and the retriever build/run under the same folder. I think this pathway vectorstore should be placed under `comps/vectorstore`.

Spycsh: Please also add the tests. Please check e.g. https://github.com/berkecanrizai/GenAIComps/blob/main/tests/test_retrievers_langchain_redis.sh

XuhuiRen, line 1: Please keep the same naming style in OPEA according to the code functionality.

comps/retrievers/langchain/pathway/data/nke-10k-2023.pdf
berkecanrizai: Hey, the Pathway vector store doesn't exactly work like the other Vector DBs. 
Vector store "listens" to the data sources registered during the construction (ie. `data/` folder in this case). We don't allow "upsert" or "chunk add" operations during runtime (though, it is possible to add files to existing sources and have them indexed). 

All of the steps (parsing, chunking, embedding, and indexing) are handled in the vector store itself.

To make it compatible with `prepare_doc_...` microservice, I could copy the files to the folder that was registered to Pathway, but that means this dataprep microservice will have to know which folders were registered to the vector store instance. Do you have any suggestions or ideas on how we could "sidestep" this?

comps/retrievers/langchain/pathway/pathway_vs.py
berkecanrizai: I think it should be ok now

berkecanrizai: Hey, just moved the vector store (docker + config file) to `comps/vectorstore`.
Also introduced the retriever tests.
What do you think about the discussion above on @lvliang-intel 's comment?

comps/vectorstores/langchain/pathway/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    libreoffice \
+    libmagic-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+RUN pip install -U pathway[xpack-llm]
+RUN pip install "unstructured[all-docs] >= 0.10.28, < 0.15"
+RUN pip install langchain_openai
+RUN pip install sentence_transformers
```
ashahba, line 20: Please create a `requirements.txt` file within `comps/vectorstores/langchain/pathway` folder and add the all `pip` dependencies there.
So you'll have something line this:
```
langchain_openai
pathway[xpack-llm]
sentence_transformers
unstructured[all-docs] '>=0.10.28,< 0.15'
```
and then update your Dockerfile this way:
```
COPY requirements.txt /app/
 
RUN pip install --no-cache-dir -r requirements.txt

COPY vectorstore_pathway.py /app/ 
```

comps/vectorstores/langchain/pathway/vectorstore_pathway.py
```diff
@@ -0,0 +1,59 @@
...
+from langchain_community.embeddings import HuggingFaceBgeEmbeddings, HuggingFaceHubEmbeddings
+from pathway.xpacks.llm.parsers import ParseUnstructured
+from pathway.xpacks.llm.vector_store import VectorStoreServer
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s %(name)s %(levelname)s %(message)s",
+    datefmt="%Y-%m-%d %H:%M:%S",
+)
+
```
ashahba, line 18: I'd add this comment line here:
```
# This creates a Pathway connector that tracks all the files in the 'data' directory
```

tests/test_retrievers_langchain_pathway.sh
```diff
@@ -0,0 +1,117 @@
...
+
+WORKPATH=$(dirname "$PWD")
+LOG_PATH="$WORKPATH/tests"
+ip_address=$(hostname -I | awk '{print $1}')
+function build_docker_images() {
+    cd $WORKPATH
+
+    cd comps/vectorstores/langchain/pathway
+
+    docker build -t vectorstore-pathway .
```
ashahba, line 15: We still need build-args for proxies here:
```
--build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy
```

comps/vectorstores/langchain/pathway/README.md
```diff
@@ -0,0 +1,84 @@
...
+- [embedders docs](https://pathway.com/developers/api-docs/pathway-xpacks-llm/embedders)
+
+## Building and running
+
+Build the Docker and run the Pathway Vector Store:
+
+```bash
+cd comps/vectorstores/langchain/pathway
+
+docker build -t vectorstore-pathway .
```
ashahba, line 62: We need build-args for proxies:
```
--build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy
```

comps/retrievers/langchain/pathway/requirements.txt
```diff
@@ -0,0 +1,13 @@
+docarray[full]
+fastapi
+frontend==0.0.3
+huggingface_hub
+langchain_community == 0.2.0
+langsmith
```
ashahba, line 6: I'd remove `langsmith` and add it to `requirements-runtime.txt` and run it through `entrypoint.sh` in your Docker.
Please see here and how it's done and let me know if you have any questions:

https://github.com/opea-project/GenAIComps/tree/main/comps/llms/summarization/tgi
https://github.com/opea-project/GenAIComps/blob/main/comps/llms/summarization/tgi/requirements.txt
https://github.com/opea-project/GenAIComps/blob/main/comps/llms/summarization/tgi/requirements-runtime.txt
https://github.com/opea-project/GenAIComps/blob/main/comps/llms/summarization/tgi/entrypoint.sh
https://github.com/opea-project/GenAIComps/blob/main/comps/llms/summarization/tgi/Dockerfile#L27

comps/retrievers/langchain/pathway/docker/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN pip install --no-cache-dir --upgrade pip && \
+    if [ ${ARCH} = "cpu" ]; then pip install torch torchvision --index-url https://download.pytorch.org/whl/cpu; fi && \
+    pip install --no-cache-dir -r /home/user/comps/retrievers/langchain/pathway/requirements.txt
+
+ENV PYTHONPATH=$PYTHONPATH:/home/user
+
+WORKDIR /home/user/comps/retrievers/langchain/pathway
+
+ENTRYPOINT ["python", "retriever_pathway.py"]
```
ashahba, line 30: This line would be changed to:
```
ENTRYPOINT ["bash", "entrypoint.sh"]
```

ashahba (changes requested):
Thanks @berkecanrizai for the contribution.
Most of my change requests are repeated within the docs, so it should be easy to make the changes. However if you need any help, please feel free to post here.

Thanks.

berkecanrizai:
> Thanks @berkecanrizai for the contribution. Most of my change requests are repeated within the docs, so it should be easy to make the changes. However if you need any help, please feel free to post here.
> 
> Thanks.

Hey, I think I resolved your comments above, thanks for the clear instructions!

ftian1 (approved):

comps/vectorstores/langchain/pathway/README.md
```diff
@@ -0,0 +1,84 @@
...
+- [embedders docs](https://pathway.com/developers/api-docs/pathway-xpacks-llm/embedders)
+
+## Building and running
+
+Build the Docker and run the Pathway Vector Store:
+
+```bash
+cd comps/vectorstores/langchain/pathway
+
+docker build --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy -t vectorstore-pathway .
```
chensuyue, line 62: The formal opea image name is like `opea/vectorstore-pathway:latest`.

tests/test_retrievers_langchain_pathway.sh
```diff
@@ -0,0 +1,117 @@
...
+function build_docker_images() {
+    cd $WORKPATH
+
+    cd comps/vectorstores/langchain/pathway
+
+    docker build --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy -t vectorstore-pathway .
+
+    cd $WORKPATH
+    # --no-cache
+    docker build -t opea/retriever-pathway:latest --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy -f comps/retrievers/langchain/pathway/docker/Dockerfile .
```
chensuyue, line 19: Please use `--no-cache` in the test, and use comps as image tag, `opea/retriever-pathway:comps`.

tests/test_retrievers_langchain_pathway.sh
```diff
@@ -0,0 +1,117 @@
...
+    export PATHWAY_HOST="0.0.0.0"
+    export PATHWAY_PORT=5432
+
+    docker run -d --name="test-comps-vectorstore-pathway" -e PATHWAY_HOST=${PATHWAY_HOST} -e PATHWAY_PORT=${PATHWAY_PORT} -e TEI_EMBEDDING_ENDPOINT=${TEI_EMBEDDING_ENDPOINT} -e http_proxy=$http_proxy -e https_proxy=$https_proxy -v $WORKPATH/comps/vectorstores/langchain/pathway/README.md:/app/data/README.md -p ${PATHWAY_PORT}:${PATHWAY_PORT} --network="host" vectorstore-pathway
+
+    sleep 45s
+
+    export PATHWAY_HOST=$ip_address  # needed in order to reach to vector store
+
+    docker run -d --name="test-retriever-pathway" -p 5009:7000 -e PATHWAY_HOST=${PATHWAY_HOST} -e PATHWAY_PORT=${PATHWAY_PORT} -e http_proxy=$http_proxy -e https_proxy=$https_proxy opea/retriever-pathway:latest
```
chensuyue, line 59: Please use `test-comps-retriever-pathway`, so the workflow can clean up all the containers even the test failed.

tests/test_retrievers_langchain_pathway.sh
```diff
@@ -0,0 +1,117 @@
...
+    else
+        echo "Result wrong. Received was $result"
+        docker logs test-comps-vectorstore-pathway >> ${LOG_PATH}/vectorstore-pathway.log
+        docker logs test-comps-retriever-tei-endpoint >> ${LOG_PATH}/tei-endpoint.log
+        docker logs test-retriever-pathway >> ${LOG_PATH}/retriever-pathway.log
+        exit 1
+    fi
+}
+
+function stop_docker() {
```
chensuyue, line 86:
```
cid=$(docker ps -aq --filter "name=test-comps*")
if [[ ! -z "$cid" ]]; then docker stop $cid && docker rm $cid && sleep 1s; fi
```
unify the container name, then those 2 sentence is enough.

tests/test_retrievers_langchain_pathway.sh
```diff
@@ -0,0 +1,117 @@
...
+
+WORKPATH=$(dirname "$PWD")
+LOG_PATH="$WORKPATH/tests"
+ip_address=$(hostname -I | awk '{print $1}')
+function build_docker_images() {
+    cd $WORKPATH
+
+    cd comps/vectorstores/langchain/pathway
+
+    docker build --build-arg https_proxy=$https_proxy --build-arg http_proxy=$http_proxy -t vectorstore-pathway .
```
chensuyue, line 15: `vectorstore-pathway` --> `opea/vectorstore-pathway:comps`

tests/test_retrievers_langchain_pathway.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
```
chensuyue, line 1: Is it passable to separate this test into 2 scripts? Based on the CI logical, when the files under `comps/vectorstores/langchain/pathway` changed, the test `test_vectorstores_langchain_pathway.sh` will be triggered. But now there is only `test_retrievers_langchain_pathway.sh`.   
You can also copy this test scripts into `test_vectorstores_langchain_pathway.sh`.

chensuyue:
Please click Resolve conversation when it's done. 
image

comps/vectorstores/langchain/pathway/data/nke-10k-2023.pdf
XuhuiRen, line 1: this file is no more a compulsory file in retriever currently. please do not default insert this file into the database

comps/retrievers/langchain/pathway/pathway_vs.py
Spycsh: Thanks. For me I think since Pathway needs a folder during construction, just keep the sample doc there is OK.

comps/vectorstores/langchain/pathway/vectorstore_pathway.py
```diff
@@ -0,0 +1,62 @@
...
+# This creates a Pathway connector that tracks all the files in the `data/` directory.
+# Additions and modifications will be reflected on the index automatically.
+
+data = pw.io.fs.read(
```
XuhuiRen, line 22: currently, the OPEA moved all data ingestion operation into the dataprep part and not more set a default file in the retriever. it would be better to follow the same style in https://github.com/opea-project/GenAIComps/tree/main/comps/retrievers/langchain/milvus to set the retriever and ingest the document in the dataprep part. Otherwise, there would be a default document that can impact the retrieval result

ashahba (approved):
Thanks @berkecanrizai 
Nice work and LGTM!

Please make sure all other reviewers' feedbacks are resolved before merging.

comps/vectorstores/langchain/pathway/vectorstore_pathway.py
berkecanrizai: I can remove the default PDF under the `data/` folder.

By dataprep, I suppose you mean something like [this](https://github.com/opea-project/GenAIComps/blob/main/comps/dataprep/milvus/prepare_doc_milvus.py).

The issue is, Pathway vector store takes the "input sources" during the construction, and keeps listening to these sources. Your PDFs and other files that are added to these folders are automatically indexed by Pathway (with the built-in parser, splitter, embedder). Hence, we don't have chunk `upsert` type of interface for the DB.
We explain this on our docs, I also tried to explain it in the `vectorstore_pathway.py`.

Do you think we could make an exception in this case?

A "work around" could be to have simple dataprep service that simply copies files into some folder Pathway is listening to, but, it is not a good solution since the Pathway input folders are configurable by the user.

tests/test_retrievers_langchain_pathway.sh
berkecanrizai: I created a new test for it. It's just a trimmed version of this retriever test without the `retriever` microservice part.

berkecanrizai: In addition, I iterated over your other comments, thanks. 
If this is ok now, I can also resolve this.

comps/vectorstores/langchain/pathway/vectorstore_pathway.py
lvliang-intel: Hi @berkecanrizai,

If users want to add or delete a file from the Pathway vectorstore, they just need to modify the 'data/' folder, right? If that's the case, we should introduce an environment variable to represent the vector store path. The `dataprep` microservice should operate exclusively on this path, saving new files to it or deleting files from it.

berkecanrizai: Yes, but that's not the whole picture. As you may see on the `pw.io.fs` line, users can add any input connectors `pw.io` as sources. Some connectors include local folders, Google Drive, SharePoint, etc. 

I have included the `data/` folder as an example in this case. 
Yes, a solution could be to have the `dataprep` microservice to save files to this `data/` folder. However, this may not be ideal since users can modify the `"sources"` config.
But surely this will work. 

In the end, if there is no option to exclude the Pathway from the other vector DBs in terms of the `dataprep` implementation, I can implement it as discussed above.

lvliang-intel: We can merge it firstly as vector database part, and we should also consider `dataprep` part for the whole pipeline.

PR #392, opened 2024-08-02
align VLLM micro-service output format with UI

xiguiw:
VLLM misco-service output format align with UI as well as that of LLM micro-service TGI implementation.

## Description

## Issues

## Type of change

micros service TGI and VLLM output format are not align.
VLLM output is not supported and displayed on UI.

TGI micro-service output format:
```
105 data: b' achieved'
106
107 data: b' record'
108
109 data: b' Rev'
```

VLLM micro-service output foramt:
```
 38 data: @#$achieved
 39
 40 data: @#$record
 41
 42 data: @#$Rev
```

## Dependencies

No

## Tests
Verified the fix the response can be output at UI.
Without fix, there are No output displayed at UI.


Spycsh:
@xiguiw , UI should work. I am curious which UI you are using? Regarding to alignment with vLLM, we currently have another PR that align in/output with OpenAI format. cc: @lkk12014402 @WenjiaoYue

xiguiw:
> @xiguiw , UI should work. I am curious which UI you are using? Regarding to alignment with vLLM, we currently have another PR that align in/output with OpenAI format. cc: @lkk12014402 @WenjiaoYue

@Spycsh 
I am using the following two UIs. They both work well with TGI.
For VLLM, UI accept input but no output response.
```
 opea/chatqna-conversation-ui:latest 
 opea/chatqna-ui:latest   
```

Spycsh (approved):

lvliang-intel (approved):

Issue #395, opened 2024-08-02
Add Versioning to Package Requirements / fastAPI 1.112 support


sunnstix:
Currently, the Gen AI components repository hardly applies any version guidelines in any of the `requirements.txt` files. The recent release of `fastapi[standard]` in version 1.112 (see [here](https://github.com/fastapi/fastapi/releases/tag/0.112.0)) is a breaking change as now many microservices will no longer have version 1.111 dependencies such as `uvicorn` automatically installed.

All requirements.txt files should at least have lower bounds in place and should fix the `fastapi` dependency changes.

PR #397, opened 2024-08-02
Quickfix: Replace fastapi dep with fastapi[standard]

sunnstix:
## Description

`fastapi` is now an alias for `fastapi-slim` as per the release of [version 1.112](https://github.com/fastapi/fastapi/releases/tag/0.112.0). 

> Before this, fastapi would include the standard dependencies, with Uvicorn and the fastapi-cli, etc.
And fastapi-slim would not include those standard dependencies.
Now fastapi doesn't include those standard dependencies unless you install with pip install "fastapi[standard]".

## Issues

#395 

## Type of change

## Dependencies

N/A

## Tests

N/A

### DIFF
--- a/comps/agent/langchain/requirements.txt
+++ b/comps/agent/langchain/requirements.txt
@@ -3,7 +3,7 @@ docarray[full]
 
 #used by tools
 duckduckgo-search
-fastapi
+fastapi[standard]
 huggingface_hub
 langchain #==0.1.12
 langchain-huggingface
--- a/comps/asr/requirements.txt
+++ b/comps/asr/requirements.txt
@@ -1,6 +1,6 @@
 datasets
 docarray[full]
-fastapi
+fastapi[standard]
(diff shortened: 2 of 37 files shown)


chensuyue:
Please fix the conflict.

lvliang-intel (approved):

chensuyue:
@sunnstix please fix the conflict.

chensuyue (approved):

chensuyue:
Please fix the conflict, so I can merge this PR. @sunnstix

sunnstix:
@chensuyue I would argue that this change no longer needs to be implemented with the new inclusion of dependencies such as uvicorn independently. The standard fastapi package introduces dependencies outside of just uvicorn that are unnecessary and if we can manually install all the needed dependencies while using the lightweight version of fastapi, I think  that is better. Feel free to close this PR if you agree.

PR #447, opened 2024-08-09
LLM micro service input data does not have input model name

xiguiw:
vLLM and oLLAMa get the model name from the envrionment

### DIFF
--- a/comps/llms/text-generation/ollama/llm.py
+++ b/comps/llms/text-generation/ollama/llm.py
@@ -21,7 +21,7 @@ from comps import GeneratedDoc, LLMParamsDoc, ServiceType, opea_microservices, r
 def llm_generate(input: LLMParamsDoc):
     ollama = Ollama(
         base_url=ollama_endpoint,
-        model=input.model,
+        model=input.model if input.model else model_name,
         num_predict=input.max_new_tokens,
         top_k=input.top_k,
         top_p=input.top_p,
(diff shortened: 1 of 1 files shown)


lvliang-intel (approved):

Spycsh:
@xiguiw why OLLAMA needs this? https://github.com/xiguiw/GenAIComps/blob/main/comps/cores/proto/docarray.py#L81 here has the input model name instead of using another env variable

comps/llms/text-generation/ollama/llm.py
```diff
@@ -49,4 +49,5 @@ async def stream_generator():
 
 if __name__ == "__main__":
     ollama_endpoint = os.getenv("OLLAMA_ENDPOINT", "http://localhost:11434")
+    model_name = os.getenv("OLLAMA_MODEL", "meta-llama/Meta-Llama-3-8B-Instruct")
```
Spycsh, line 52: Why meta-llama/Meta-Llama-3-8B-Instruct?

xiguiw: It's default value. In case OLLAMA_MODEL is not set, then it is 'meta-llama/Meta-Llama-3-8B-Instruct'

Spycsh: model=input.model if input.model else model_name

PR #508, opened 2024-08-16
ollama: Update curl proxy.

feng-intel:
## Description
add noproxy in curl command.

## Issues
Need unset http_proxy and https_proxy, or curl failed.

## Type of change
doc

## Dependencies
No

## Tests
Pass

### DIFF
--- a/comps/llms/text-generation/ollama/README.md
+++ b/comps/llms/text-generation/ollama/README.md
@@ -40,7 +40,7 @@ All of your local models are automatically served on localhost:11434. Run ollama
 Send an application/json request to the API endpoint of Ollama to interact.
 
 ```bash
-curl http://localhost:11434/api/generate -d '{
+curl --noproxy "*" http://localhost:11434/api/generate -d '{
   "model": "llama3",
   "prompt":"Why is the sky blue?"
 }'


mkbhanda (approved):
I am puzzled why/when this proxy stuff changed. From India with proxy on my git clone commands stopped working! Is this related to the new auto VPN?

mkbhanda:
@feng-intel  DCO missing! docker file path change.

lvliang-intel (approved):

comps/llms/text-generation/ollama/README.md
```diff
@@ -40,7 +40,7 @@ All of your local models are automatically served on localhost:11434. Run ollama
 Send an application/json request to the API endpoint of Ollama to interact.
 
 ```bash
-curl http://localhost:11434/api/generate -d '{
+curl --noproxy "*" http://localhost:11434/api/generate -d '{
```
ftian1, line 43: shall we use env variable HTTP_PROXY, HTTPS_PROXY, NO_PROXY to control the behavior but not in curl cmd self?

if it doesn't work, another way may be asking user to edit .curlrc with their own proxy config. so the cmd line here is general for all cases.

Spycsh (approved):

comps/llms/text-generation/ollama/README.md
mkbhanda: Personally I like the idea of editting .curlrc file. Even for GitHub commands like pull etc, I needed different settings depending on where I worked from. So let us just provide documentation, "assuming you have set your .curlrc to reflect proxy settings that you may need in your environment. Example settings are: .."

PR #573, opened 2024-08-28
remove finetuning models limitation.

lkk12014402:
## Description

support more models for finetuning


lkk12014402:
### dashboard

image

comps/finetuning/handlers.py
```diff
@@ -62,18 +60,12 @@ def handle_create_finetuning_jobs(request: FineTuningJobsRequest, background_tas
...
 
-    model_config_file = MODEL_CONFIG_FILE_MAP.get(base_model)
-    if not model_config_file:
-        raise HTTPException(status_code=404, detail=f"Base model '{base_model}' not supported!")
-
     if not os.path.exists(train_file_path):
         raise HTTPException(status_code=404, detail=f"Training file '{train_file}' not found!")
 
-    with open(model_config_file) as f:
-        finetune_config = parse_yaml_raw_as(FinetuneConfig, f)
```
XinyuYe-Intel, line 73: It would lose many flexibility by not using yaml. 
For example, how do you use lora? looks like lora is not used by default. 
Also, finetuning of rerank and embedding model is somewhat different with the default config for instruction tuning, thus need pass in some parameters different with the default ones, how to handle such case? parameters of `FineTuningJobsRequest` seems not enough to cover most circumstances.

lkk12014402: 1. using yaml doesn't have flexibility for different models, so I remove them.

2. the lora setting is same with yaml (I change `lora_config: Optional[LoraConfig] = None` to lora_config: Optional[LoraConfig] = LoraConfig()`)

4. I will extend FineTuningJobsRequest to be able to customize `GeneralConfig`, `DatasetConfig`, `TrainingConfig`

XinyuYe-Intel: Using yaml in current way of binding yaml to model indeed does not have flexibility, but when passing yaml to FineTuningJobsRequest, I think it will have much flexibility.

lkk12014402: so we should extend FineTuningJobsRequest to be able to customize GeneralConfig, DatasetConfig, TrainingConfig? do you agree?

XinyuYe-Intel: Yes, either add parameters directly in FineTuningJobsRequest or just add an argument for yaml.

lkk12014402: ok, I will add parameters directly in FineTuningJobsRequest

lvliang-intel (approved):

lkk12014402:
hi, @XinyuYe-Intel  please review again. I add `FineTuningParams` for parameters customization.

kevinintel (approved):

tests/test_ray_finetuning_on_cpu.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
chensuyue, line 1: The test name must be `test_finetuning.sh`, otherwise test won't be triggered when there is no test scripts change.

WenjiaoYue (approved):

Zhenzhong1 (approved):

XinyuYe-Intel (approved):

PR #660, opened 2024-09-10
fix ragagent text generator bug

minmin-intel:
## Description

The "query_time" needs to be initialized with an empty string. Otherwise the agent will error.

## Issues

na

## Type of change

## Dependencies

no

## Tests

unit test


xuechendi (approved):
LGTM

hshen14 (approved):

PR #730, opened 2024-09-24
Update RetrievalTool Gateway to allow user setting retriever/reranker params

minmin-intel:
## Description

Previously user can only pass in query to the retrieval tool and there is no way to set retriever/reranker params with requests to the gateway. This PR is to expose these params to users and allow users to set retriever/reranker params with the requests to the gateway.

## Issues

`n/a`.

## Type of change

## Dependencies

No newly introduced 3rd party dependency.

## Tests

DocIndexRetriever example test


comps/cores/mega/gateway.py
```diff
@@ -622,31 +622,41 @@ def __init__(self, megaservice, host="0.0.0.0", port=8889):
...
-        for key, TypeClass in zip(["text", "input", "input"], [TextDoc, EmbeddingRequest, ChatCompletionRequest]):
-            query = parser_input(data, TypeClass, key)
-            if query is not None:
-                break
-        if query is None:
-            raise ValueError(f"Unknown request type: {data}")
-        result_dict, runtime_graph = await self.megaservice.schedule(initial_inputs={"text": query})
+        chat_request = ChatCompletionRequest.parse_obj(data)
+        query = chat_request.input
+        print("Query: ", query)
```
xuechendi, line 633: need remove print here

comps/cores/mega/gateway.py
```diff
@@ -622,31 +622,41 @@ def __init__(self, megaservice, host="0.0.0.0", port=8889):
             host,
             port,
             str(MegaServiceEndpoint.RETRIEVALTOOL),
-            Union[TextDoc, EmbeddingRequest, ChatCompletionRequest],  # ChatCompletionRequest,
-            Union[RerankedDoc, LLMParamsDoc],  # ChatCompletionResponse
+            ChatCompletionRequest,
```
xuechendi, line 625: Are we not expecting TextDoc / EmbeddingRequest in future retrieval request? @ftian1 @lvliang-intel

comps/cores/mega/gateway.py
```diff
@@ -622,31 +622,41 @@ def __init__(self, megaservice, host="0.0.0.0", port=8889):
...
+            distance_threshold=chat_request.distance_threshold if chat_request.distance_threshold else None,
+            fetch_k=chat_request.fetch_k if chat_request.fetch_k else 20,
+            lambda_mult=chat_request.lambda_mult if chat_request.lambda_mult else 0.5,
+            score_threshold=chat_request.score_threshold if chat_request.score_threshold else 0.2,
+        )
+        reranker_parameters = RerankerParms(
+            top_n=chat_request.top_n if chat_request.top_n else 1,
+        )
+
+        print("Retriever parameters:\n", retriever_parameters)
```
xuechendi, line 647: remove print

comps/cores/mega/gateway.py
```diff
@@ -622,31 +622,41 @@ def __init__(self, megaservice, host="0.0.0.0", port=8889):
...
+        result_dict, runtime_graph = await self.megaservice.schedule(
+            initial_inputs={"messages": "", "input": query, "k": chat_request.k, "top_n": chat_request.top_n},
+            retriever_parameters=retriever_parameters,
+            reranker_parameters=reranker_parameters,
+        )
+
         last_node = runtime_graph.all_leaves()[-1]
         response = result_dict[last_node]
-        print("response is ", response)
+        print("Response:\n", response)
```
xuechendi, line 659: remove print

comps/cores/mega/gateway.py
minmin-intel: Good catch. I think we should also support TextDoc and EmbeddingRequest inputs. Let me add them back in.

comps/cores/mega/gateway.py
minmin-intel: Done

comps/cores/mega/gateway.py
minmin-intel: Done.

comps/cores/mega/gateway.py
minmin-intel: @xuechendi  I added the other input types back. Please review.

xuechendi (approved):
Codes Looks good to me
My only concern is that it looks like this PR is not covered by any UT?

xuechendi:
@chensuyue , is there any UT will cover gateway code change in GenAIComp?

minmin-intel:
> Codes Looks good to me My only concern is that it looks like this PR is not covered by any UT?

Looks like they pick ChatQnA example as UT. But I agree that the code changes in RetrievalToolGate was not tested in CICD. @chensuyue Can we add DocIndexRetriever example in the Example Test?

lvliang-intel (approved):

PR #750, opened 2024-09-27
Fix tool_render with direct import from pydantic

minmin-intel:
## Description

latest langchain structuredtool needs to import from pydantic. Old tool renderer gives error with tool args. Fix tool render by directly import from pydantic

## Issues

 `n/a`.

## Type of change

## Dependencies

No newly introduced 3rd party dependency if exists.

## Tests

Unit test


comps/agent/langchain/src/strategy/ragagent/planner.py
```diff
@@ -8,13 +8,13 @@
...
 from langchain_core.output_parsers.openai_tools import PydanticToolsParser
 from langchain_core.prompts import PromptTemplate
-from langchain_core.pydantic_v1 import BaseModel, Field
 from langchain_huggingface import ChatHuggingFace, HuggingFaceEndpoint
 from langchain_openai import ChatOpenAI
 from langgraph.checkpoint.memory import MemorySaver
 from langgraph.graph import END, START, StateGraph
 from langgraph.graph.message import add_messages
 from langgraph.prebuilt import ToolNode, tools_condition
+from pydantic import BaseModel, Field
```
xuechendi, line 17: any specific reason for this change?

xuechendi (approved):
LGTM

xuechendi:
@minmin-intel , may you also update the langchain version who need : from pydantic import BaseModel, Field; So we can avoid install old version which breaks the workflow

minmin-intel:
> @minmin-intel , may you also update the langchain version who need : from pydantic import BaseModel, Field; So we can avoid install old version which breaks the workflow

We are installing latest langchain in requirement.txt 
image
Did you want to fix the version? @xuechendi

lvliang-intel (approved):

PR #774, opened 2024-10-09
add dynamic batching embedding/reranking

Spycsh:
## Description

To solve the issue that multiple "small" GenAI services cannot be launched on one Gaudi card, here we draft one implementation of replacing TEI logics with local inference using optimum habana + dynamic batching. This requires the following functionalities:

- request buffering
- HPU batch inference on embedding/reranking

## Issues

n/a

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)
- [ ] Others (enhancement, documentation, validation, etc.)

## Dependencies

Under implementation

## Tests

Under implementation

PR #794, opened 2024-10-16
Add image2image microservice

XinyuYe-Intel:
## Description

Add image2image microservice.

## Issues

`n/a`.

## Type of change

## Dependencies

## Tests


lvliang-intel (approved):

comps/image2image/image2image.py
```diff
@@ -0,0 +1,117 @@
...
+)
+
+logger = CustomLogger("image2image")
+pipe = None
+args = None
+initialization_lock = threading.Lock()
+initialized = False
+
+
+def initialize():
```
Spycsh, line 32: Just curious why you use a lock instead of initialize the models in `__main__`. Does the subprocess hang? If so, you can consider to also use `opea_microservices["opea_service@image2image"].start(in_single_process=True)` in the future. For now I think it looks good to me.

Spycsh (approved):

comps/image2image/image2image.py
XinyuYe-Intel: Yes, model initialized in `__main__` will not in the same process as microservice, which would cause trouble in Gaudi.
Looks like `MicroService.start` does not have `in_single_process` argument now, what is this argument meant to do?

Spycsh: Oh, I see. https://github.com/opea-project/GenAIComps/pull/774/files#diff-31e12d9abc1b692d578d4b76013e5bdf0c367cbe86904c79be8594653c9ede36R172 still has not been merged, in which I expose `in_single_process=True` which can start the microservice in the same process to prevent workload from HPU segmentation fault issue.

PR #798, opened 2024-10-16
[Agent] support custom prompt

xuechendi:
## Description

This PR is to support user provided custom_prompt when launching agent.
We hold an assumption that the new prompt should be as same name as default one. => Then we will replace exiting one with custom prompt to fit better for certain use cases.

## Type of change

## Tests

covered with existing test


xuechendi:
added a new argument - custom_prompt
image

And anything defines in custom_prompt.py will replace existing ones.

For Example:

I did a print in react_langgraph class
image

and from container log, we can see now it is using custom_prompt
image

comps/agent/langchain/src/utils.py
```diff
@@ -122,6 +123,23 @@ def has_multi_tool_inputs(tools):
     return ret
 
 
+def load_python_prompt(file_dir_path: str):
```
minmin-intel, line 126: What is this function used for?

xuechendi: This function is used for loading modules from python file.

minmin-intel: I don't see this function used anywhere else?

comps/agent/langchain/src/agent.py
```diff
@@ -1,30 +1,37 @@
 # Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
+from .utils import load_python_prompt
 
 
 def instantiate_agent(args, strategy="react_langchain", with_memory=False):
+    if args.custom_prompt is not None:
+        print(f">>>>>> custom_prompt enabled, {args.custom_prompt}")
+        custom_prompt = load_python_prompt(args.custom_prompt)
```
xuechendi, line 9: @minmin, this function will be used here, it is transparent to strategy developers

minmin-intel (approved):

PR #826, opened 2024-10-25
fix pydantic version in agent component

minmin-intel:
## Description

change pydantic import in utils.py to fix the tool_renderer error.

## Issues
`n/a`.

## Type of change

## Dependencies

No newly introduced 3rd party dependency.

## Tests

Unit test

### DIFF
--- a/comps/agent/langchain/src/tools.py
+++ b/comps/agent/langchain/src/tools.py
@@ -8,7 +8,7 @@ import os
 import yaml
 from langchain.tools import BaseTool, StructuredTool
 from langchain_community.agent_toolkits.load_tools import load_tools
-from pydantic.v1 import BaseModel, Field, create_model
+from pydantic import BaseModel, Field, create_model
 
 
 def generate_request_function(url):


xuechendi (approved):
LGTM

xuechendi:
@lkk12014402 , please help to approve

dcmiddle:
It looks like this version pinning went in less than a week ago.
Is reverting it just going to bring up the same conflict?
https://github.com/opea-project/GenAIComps/pull/782

minmin-intel:
> It looks like this version pinning went in less than a week ago. Is reverting it just going to bring up the same conflict? #782

@dcmiddle Hi, the #782 PR was put in as a quick fix for [this PR](https://github.com/opea-project/GenAIExamples/pull/892) in GenAIExamples. Back then the unit tests for agent comp passed. But recently, we got a couple of reports from users that pydantic.v1 gave them errors, changing to import from pydantic fixed their errors. So we reached out to the author of PR892 and he was able to use pydantic instead of pydantic.v1 by changing his code. So we decided to change back to import from pydantic.

lvliang-intel (approved):

dcmiddle:
Ok, thanks for the added background. Sounds like @JoshuaL3000 should be good with pydantic v2 then.

PR #835, opened 2024-10-30
Multiple models support for LLM TGI

sgurunat:
Description
To support multiple llm models for ChatQnA, the changes are incorporated into llms TGI text-generation. Multiple models can be provided in model_configs.json file which will be loaded into MODEL_CONFIGS environment variable.

Type of change
 New feature (non-breaking change which adds new functionality)
##Changes
To support this the model parameter has been added in the ChatQnAGateway and LLMParams from gateway.py and docarray.py respectively.
Added load_model_configs method in utils.py to validate all the required fields ( 'model_name', 'displayName', 'endpoint', 'minToken', 'maxToken') and then load the configurations. This is added in utils so that it can be reused.
Updated llm.py from llms text-generation tgi to support multiple models and transfer the call to right endpoint.
Updated the template.py file from llms text-generation tgi to have new template for models meta-llama/Meta-Llama-3.1-70B-Instruct" and "meta-llama/Meta-Llama-3.1-8B-Instruct"


comps/llms/text-generation/tgi/llm.py
```diff
@@ -45,6 +56,17 @@
...
+
+    if input.model and MODEL_CONFIGS and configs_map:
+        if configs_map.get(input.model):
+            config = configs_map.get(input.model)
+            llm_endpoint = config.get("endpoint")
+        else:
+            logger.error(f"Input model {input.model} not present in model_configs")
+            raise ConfigError(f"Input model {input.model} not present in model_configs")
+
+    llm = AsyncInferenceClient(model=llm_endpoint, timeout=600)
```
lvliang-intel, line 68: Please check the CI test error, looks like llm_endpoint is empty string for this case, please fix it.
image

sgurunat: Thanks for the review. Fixed it.

ftian1:
I am confused by this PR. why we want user to pass a model_config to support different models? each OPEA microservice's instance will only support 1 model during deployment, The model_id will not be changed. 

I don't think this is right requirement to support changing different models during inference request.

comps/llms/text-generation/tgi/llm.py
```diff
@@ -45,6 +56,18 @@
 async def llm_generate(input: Union[LLMParamsDoc, ChatCompletionRequest, SearchedDoc]):
     if logflag:
         logger.info(input)
+
+    llm_endpoint = DEFAULT_ENDPOINT
+    if input.model and MODEL_CONFIGS and configs_map:
```
letonghan, line 61: Can you simplify these code?
Seems like MODEL_CONFIGS is not needed here, and llm_endpoint can get using `configs_map.get(input.model).get("endpoint")`.
The exception should be catched and raised in a 'try except' block.

comps/llms/text-generation/tgi/llm.py
```diff
@@ -22,16 +22,27 @@
...
-    model=llm_endpoint,
-    timeout=600,
-)
+# Environment variables
+MODEL_CONFIGS = os.getenv("MODEL_CONFIGS")
+DEFAULT_ENDPOINT = os.getenv("TGI_LLM_ENDPOINT", "http://localhost:8080")
+
+# Extract the model endpoint
+llm_endpoint = "http://localhost:8080"
+configs_map = {}
```
letonghan, line 37: To keep the code clean, the codes that get 'llm_endpoint' here should be wrapped into a funtion, and called when initializing llm object.

sgurunat: Have created function to get llm_endpoint and it is being called in llm_generate function. I have kept the load_model_configs outside itself because its a static content and one time load during bringing up the service.

comps/llms/text-generation/tgi/llm.py
sgurunat: Simplified this code as well

ftian1 (approved):

ftian1:
after discussion, we think having such model switch capability in s/w layer is right way to go. just the model_config format may be able to revise for simplicity. it could be done in future PR.

lvliang-intel:
@sgurunat, 
please resolve the conflict and then we can merge the PR.

lvliang-intel (approved):

lvliang-intel:
@sgurunat,
please resolve the code conflict.

ENCODERS09 (approved):

sgurunat:
@lvliang-intel - resolved the merge conflicts

PR #889, opened 2024-11-11
Minor simplication to ServiceOrchestrator code

eero-t:
## Description

`dump_outputs()` method in `ServiceOrchestrator`:
* Is not real method (does not use `self`)
* Adds a member to a dict instead of "dump"ing (dropping or outputting) something
* Obfuscates how `schedule()` method return value is constructed, and
* Makes calling lines unnecessary longer

So this dumps it.  Similar method in `ServiceOrchestratorWithYaml` is more reasonable[1], but is dumped for consistency.

This also applies `pylint` simplification suggestion to the code.

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

`pylint` + manual testing.

[1] IMHO `add_result()` would have been reasonable name for the `ServiceOrchestratorWithYaml` method.


eero-t:
Added also simplication suggested by pylint + rebased to main.

PR #915, opened 2024-11-19
fix retriever and reranker to process chat completion request

minmin-intel:
## Description
When testing DocIndexRetriever with user input k and top_n as ChatCompletionRequest, I identified two errors:
1. retriever redis: previously if input is chat completion request, since the embedding is EmbeddingResponse, it cannot be processed by the vector db, so need to add additional steps to get the embedding vectors out from the EmbeddingResponse.
2. reranker tei: previously input_type is fixed as SearchedDoc, so if the input is ChatCompletionRequest, the microservice will return error 422 unprocessable entity.

As a result of these 2 errors, the DocIndexRetriever fails when user send ChatCompletionRequest.

This PR fixes the issues above. 

## Issues

## Type of change

## Dependencies

## Tests

retriever and reranker unit tests, DocIndexRetriever example tests.

### DIFF
--- a/comps/cores/mega/gateway.py
+++ b/comps/cores/mega/gateway.py
@@ -773,7 +773,7 @@ class RetrievalToolGateway(Gateway):
             host,
             port,
             str(MegaServiceEndpoint.RETRIEVALTOOL),
-            Union[TextDoc, EmbeddingRequest, ChatCompletionRequest],
+            Union[TextDoc, ChatCompletionRequest],
             Union[RerankedDoc, LLMParamsDoc],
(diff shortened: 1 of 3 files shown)


lkk12014402 (approved):

chensuyue:
Bug fix for DocIndexRetriever failed in v1.1 release test, https://github.com/opea-project/GenAIExamples/actions/runs/11876357287/job/33118149000

chensuyue (approved):

lvliang-intel (approved):

PR #967, opened 2024-12-03
Add helm starter chart for developing helm chart

lianhao:
## Description

Helm starter chart pack can be used to create helm chart for new services.

## Issues

`n/a`.

## Type of change

## Dependencies

## Tests

### DIFF
--- a/dev/helm-chart-starter/README.md
+++ b/dev/helm-chart-starter/README.md
@@ -1,0 +1,29 @@
+# OPEA <CHARTNAME> microservice
+
+Helm chart for deploying OPEA example service.
+
+## Installing the Chart
+
+To install the chart, run the following:
+
+```console
+
+```
+
+## Verify
+
+To verify the installation, run the command `kubectl get pod` to make sure all pods are runinng and in ready state.
+
+Then run the command `kubectl port-forward svc/example 8080:8080` to expose the tgi service for access.
+
+Open another terminal and run the following command to verify the service if working:
+
+```console
+
+```
+
+## Values
+
+| Key                             | Type   | Default                              | Description            |
+| ------------------------------- | ------ | ------------------------------------ | ---------------------- |
+| global.HUGGINGFACEHUB_API_TOKEN | string | `insert-your-huggingface-token-here` | Hugging Face API token |
--- a/dev/helm-chart-starter/templates/_helpers.tpl
+++ b/dev/helm-chart-starter/templates/_helpers.tpl
@@ -1,0 +1,64 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "<CHARTNAME>.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "<CHARTNAME>.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "<CHARTNAME>.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "<CHARTNAME>.labels" -}}
+helm.sh/chart: {{ include "<CHARTNAME>.chart" . }}
+{{ include "<CHARTNAME>.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "<CHARTNAME>.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "<CHARTNAME>.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "<CHARTNAME>.serviceAccountName" -}}
+{{- if .Values.global.sharedSAName }}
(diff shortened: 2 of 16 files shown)


lianhao:
Let's hold this until we've got the helm charts directory layout figured out

dev/helm-chart-starter/templates/tests/test-pod.yaml
```diff
@@ -0,0 +1,29 @@
...
+metadata:
+  name: "{{ include "<CHARTNAME>.fullname" . }}-testpod"
+  labels:
+    {{- include "<CHARTNAME>.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+spec:
+  containers:
+    - name: curl
+      image: python:3.10.14
```
eero-t, line 15: Latest Python 3.10 bug/security-fix version is 3.10.16 (and current latest Python version is 3.13.1).

dev/helm-chart-starter/values.yaml
```diff
@@ -0,0 +1,167 @@
...
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+# This is to setup the liveness and readiness probes more information can be found here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
```
eero-t, line 94:
```suggestion
# This is to setup the startup, liveness and readiness probes more information can be found here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
```

Or:
```suggestion
# Set up pod health monitoring probes, see: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
```

dev/helm-chart-starter/values.yaml
```diff
@@ -0,0 +1,167 @@
...
+    port: <CHARTNAME>
+  initialDelaySeconds: 5
+  periodSeconds: 5
+  failureThreshold: 120
+
+# This section is for setting up autoscaling more information can be found here: https://kubernetes.io/docs/concepts/workloads/autoscaling/
+# Enabling HPA will:
+# - Ignore above replicaCount, as it will be controlled by HPA
+# - Add example HPA scaling rules with thresholds suitable for Xeon deployments
+# - Require custom metrics ConfigMap available in the main application chart
```
eero-t, line 120: Fix comment (similarly to the other PR).

Because HPA can sometimes change replica counts up and down rather frequently, dev documentation (or this comment) should emphatize that service needs to handle SIGTERM and:
* stop accepting new requests
* handle _all of its buffered requests_
* terminate after tthose have been processed

To avoid k8s SIGKILL (sent to pod after specified grace period) causing request failures.

See: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

dev/helm-chart-starter/templates/horizontal-pod-autoscaler.yaml
```diff
@@ -0,0 +1,68 @@
...
+        averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+  {{- end }}
+  {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
+  - type: Resource
+    resource:
+      name: memory
+      target:
+        type: Utilization
+        averageUtilization: {{ .Values.autoscaling.targetMemoryUtilizationPercentage }}
+  {{- end }}
```
eero-t, line 34: IMHO these variable names are too long/verbose.  I'd suggest using just:
* targetCPUPercentage
* targetMemoryPercentage

dev/helm-chart-starter/values.yaml
```diff
@@ -0,0 +1,167 @@
...
+global:
+  http_proxy: ""
+  https_proxy: ""
+  no_proxy: ""
+  HUGGINGFACEHUB_API_TOKEN: "insert-your-huggingface-token-here"
+  # service account name to be shared with all parent/child charts.
+  # If set, it will overwrite serviceAccount.name.
+  # If set, and serviceAccount.create is false, it will assume this service account is already created by others.
+  sharedSAName: ""
+  # Install Prometheus serviceMonitor for service
```
eero-t, line 164:
```suggestion
  # Install Prometheus serviceMonitor for service metrics
```

dev/helm-chart-starter/templates/horizontal-pod-autoscaler.yaml
```diff
@@ -0,0 +1,68 @@
...
+      describedObject:
+        apiVersion: v1
+        # get metric for named object of given type (in same namespace)
+        kind: Service
+        name: {{ include "<CHARTNAME>.fullname" . }}
+      target:
+        type: AverageValue
+        averageValue: 15
+      metric:
+        name: {{ include "<CHARTNAME>.metricPrefix" . }}_dummy_queue_size
```
eero-t, line 47: Remove this part as this template does not set up any custom metrics for this application, and because there would be no point in doing that (custom metrics are by definition application specific).

dev/helm-chart-starter/templates/_helpers.tpl
```diff
@@ -0,0 +1,71 @@
...
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
+{{- end }}
+{{- end }}
+
+{{/*
+Convert chart name to a string suitable as metric prefix
+*/}}
+{{- define "<CHARTNAME>.metricPrefix" -}}
+{{- include "<CHARTNAME>.fullname" . | replace "-" "_" | regexFind "[a-zA-Z_:][a-zA-Z0-9_:]*" }}
```
eero-t, line 70: Remove this part as this template does not set up any custom metrics for this application.

dev/helm-chart-starter/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```console
+
+```
+
+## Values
+
+| Key                             | Type   | Default                              | Description            |
+| ------------------------------- | ------ | ------------------------------------ | ---------------------- |
+| global.HUGGINGFACEHUB_API_TOKEN | string | `insert-your-huggingface-token-here` | Hugging Face API token |
```
eero-t, line 29: I think this needs a link to a shared documentation explaining all the common keys in the values file (proxies etc). Such doc could be split to sections based on how commonly given keys need to be specified by the user.

dev/helm-chart-starter/values.yaml
```diff
@@ -0,0 +1,167 @@
...
+# Enabling HPA will:
+# - Ignore above replicaCount, as it will be controlled by HPA
+# - Add example HPA scaling rules with thresholds suitable for Xeon deployments
+# - Require custom metrics ConfigMap available in the main application chart
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 4
+  # targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
```
eero-t, line 126: Helm install fails if there are not targets.  Because these do nothing unless autoscaling is enabled, I think they can be always enabled:
```suggestion
  targetCPUPercentage: 80
  targetMemoryPercentage: 80
```

eero-t (changes requested):
I think some instructions on using the template are also needed.

It could e.g. state which classes service needs to use, for it to have `v1/health_check` endpoint for the probes, and that services always need to handle `SIGTERM`, to avoid lost requests on service scale down.

PR #1020, opened 2024-12-10
Helmchart

yongfengdu:
## Description

## Issues

## Type of change

## Dependencies

## Tests


lianhao:
@yongfengdu  could you please split this into multiple independent PRs. It's too much files to review and make my browser hangs

lianhao:
Also, gpt-sovits is one of the opea tts services, it shouldn't belong to 3rd party

mkbhanda:
Indeed smaller commits more sane! Thank you for all the hard work @yongfengdu . BTW I noticed that mongodb had a redis description but I did not get very far in my review

yongfengdu:
> Indeed smaller commits more sane! Thank you for all the hard work @yongfengdu . BTW I noticed that mongodb had a redis description but I did not get very far in my review

Good catch, the mongodb one(Chart.yaml) was copied from redis but leaked the replacement.
Yes, to use helm charts from original is a good idea, like the case of milvus. But for the 3rd party components like DB or VectorDB, it's more likely customer(Or in Example) will have their own optimized solution for production.

PR #1092, opened 2024-12-30
Create token metrics only when they are available

eero-t:
## Description

This avoids generating useless token / request histogram metrics for services that use Orchestrator class, but never call its token processing functionality.   Such dummy metrics can confuse telemetry users.

(It also helps in differentiating frontend megaservice metrics from backend megaservice ones, especially when multiple OPEA applications with wrapper microservices run in the same cluster.)

## Issues

`n/a`.

## Type of change

## Dependencies

`n/a`.

## Tests

Manual testing with latest versions, to verify that:
* services processing tokens, generate token histogram metrics
* ones not processing them, produce only pending requests gauge

### DIFF
--- a/comps/cores/mega/orchestrator.py
+++ b/comps/cores/mega/orchestrator.py
@@ -7,6 +7,7 @@ import copy
 import json
 import os
 import re
+import threading
 import time
 from typing import Dict, List
 
@@ -42,27 +43,40 @@ class OrchestratorMetrics:
 
         self.request_pending = Gauge(f"{self._prefix}_request_pending", "Count of currently pending requests (gauge)")
 
+        # locking for latency metric creation / method change
+        self._lock = threading.Lock()
+
         # Metrics related to token processing are created on demand,
         # to avoid bogus ones for services that never handle tokens
         self.first_token_latency = None
         self.inter_token_latency = None
         self.request_latency = None
 
-        # initial methods create the metric
+        # initial methods to create the metrics
         self.token_update = self._token_update_create
         self.request_update = self._request_update_create
 
     def _token_update_create(self, token_start: float, is_first: bool) -> float:
-        self.first_token_latency = Histogram(f"{self._prefix}_first_token_latency", "First token latency (histogram)")
-        self.inter_token_latency = Histogram(f"{self._prefix}_inter_token_latency", "Inter-token latency (histogram)")
-        self.token_update = self._token_update_real
+        with self._lock:
+            # in case another thread already got here
+            if self.token_update == self._token_update_create:
+                self.first_token_latency = Histogram(
+                    f"{self._prefix}_first_token_latency", "First token latency (histogram)"
+                )
+                self.inter_token_latency = Histogram(
+                    f"{self._prefix}_inter_token_latency", "Inter-token latency (histogram)"
+                )
+                self.token_update = self._token_update_real
         return self.token_update(token_start, is_first)
 
     def _request_update_create(self, req_start: float) -> None:
-        self.request_latency = Histogram(
-            f"{self._prefix}_request_latency", "Whole LLM request/reply latency (histogram)"
-        )
-        self.request_update = self._request_update_real
+        with self._lock:
+            # in case another thread already got here
+            if self.request_update == self._request_update_create:
+                self.request_latency = Histogram(
+                    f"{self._prefix}_request_latency", "Whole LLM request/reply latency (histogram)"
+                )
+                self.request_update = self._request_update_real
         self.request_update(req_start)
 
     def _token_update_real(self, token_start: float, is_first: bool) -> float:
--- a/comps/cores/mega/orchestrator.py
+++ b/comps/cores/mega/orchestrator.py
@@ -27,20 +27,45 @@ LOGFLAG = os.getenv("LOGFLAG", False)
 
 
 class OrchestratorMetrics:
-    # Because:
+    # Need an instance ID for metric prefix because:
+    # - Orchestror instances are not named
     # - CI creates several orchestrator instances
-    # - Prometheus requires metrics to be singletons
-    # - Oorchestror instances are not provided their own names
-    # Metrics are class members with "megaservice" name prefix
-    first_token_latency = Histogram("megaservice_first_token_latency", "First token latency (histogram)")
-    inter_token_latency = Histogram("megaservice_inter_token_latency", "Inter-token latency (histogram)")
-    request_latency = Histogram("megaservice_request_latency", "Whole request/reply latency (histogram)")
-    request_pending = Gauge("megaservice_request_pending", "Count of currently pending requests (gauge)")
+    # - Prometheus requires metrics (their names) to be unique
+    _instance_id = 0
 
     def __init__(self) -> None:
-        pass
+        self._instance_id += 1
+        if self._instance_id > 1:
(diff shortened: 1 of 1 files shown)


eero-t:
@Spycsh Could you review this?

eero-t:
Rebased to `main`.

comps/cores/mega/orchestrator.py
```diff
@@ -25,28 +25,42 @@
 
 
 class OrchestratorMetrics:
-    # Because:
+    # Metrics use "megaservice_" name prefix, and are class variables because:
     # - CI creates several orchestrator instances
     # - Prometheus requires metrics to be singletons
```
Spycsh, line 30: Here you mention singletons but on line https://github.com/opea-project/GenAIComps/blob/62e24c271e901a9219fcc2ff2a2b7cc7a07c6a8c/comps/cores/mega/orchestrator.py#L68 you are still using inc/dec for each single OrchestratorMetrics instance, right? Consider using `type(self).request_latency`?

Spycsh:
> This avoids generating useless token / request histogram metrics for services that use Orchestrator class, but never call its token processing functionality. Such dummy metrics can confuse telemetry users.

Why will this happen? Metrics are only updated when calling `self.metrics.pending_update`-like methods in `schedule`, right? These are all controllable code. So what you mean is that there are some other thing that can update the metrics?

comps/cores/mega/orchestrator.py
eero-t: Functionally they are identical (at least as long as class is not subclassed).

But I decided instead to use different metric prefixes for additional Metrics/Orchestrator instances.

eero-t:
I dropped pending metric doc update & rebased to main.  I'll have it in separate PR where I fix additional issues I noticed, which require pending requests metric type / name change.

eero-t:
`opea/dataprep-redis:latest` does not seem to generate `megaservice_*` metrics anymore.

Has `ServiceOrchestrator` use been dropped from backend services?

eero-t:
> I dropped pending metric doc update & rebased to main. I'll have it in separate PR where I fix additional issues I noticed, which require pending requests metric type / name change.

Could not find any good fix for it, so I just filed a ticket on it: https://github.com/opea-project/GenAIComps/issues/1121

Spycsh:
> This avoids generating useless token / request histogram metrics for services that use Orchestrator class, but never call its token processing functionality. Such dummy metrics can confuse telemetry users.
> 
> (It also helps in differentiating frontend megaservice metrics from backend megaservice ones, especially when multiple OPEA applications with wrapper microservices run in the same cluster.)

OK so what you mean it that the dummy metrics will show zeros after initialization and before the first request and users should not see wrong values of request number... But you think the k8s will scrape the metrics even there are no requests and it is resource-consuming so you decide to delay the initialization only when there are requests. I agree with this approach.

Spycsh:
> `opea/dataprep-redis:latest` does not seem to generate `megaservice_*` metrics anymore.
> 
> Has `ServiceOrchestrator` use been dropped from backend services?

dataprep microservice itself should not generate `megaservice_*` metrics. Only megaservices like `opea/chatqna` do.

eero-t:
> OK so what you mean it that the dummy metrics will show zeros after initialization and before the first request and users should not see wrong values of request number...

Technically the zero counts are not wrong, but _presence_ of token / LLM metrics is misleading for services that will **never** generate tokens (or use LLM).  That's the main reason for this PR.

> But you think the k8s will scrape the metrics even there are no requests and it is resource-consuming so you decide to delay the initialization only when there are requests. I agree with this approach.

All OPEA originated services use HttpService i.e. provide HTTP access metrics [1].  To see those, `serviceMonitor`s are installed for them when `monitoring` option is enabled in Helm charts. Meaning that any `megaservice_*` metrics they generate, will also be visible to user e.g. in Grafana.

I doubt skipping generation of extra metrics has any noticeable perf impact on the service providing the metrics (currently `serviceMonitor`s are configured to poll them at 5s interval), but every little bit can help.

Note: each Prometheus Histogram type provides about dozen different metrics, and in larger clusters, amount of metrics needs to be reduced to keep telemetry stack resource usage & perf reasonable. Telemetry stack resource usage is significant concern only when there's larger number of such pods though.

-----

[1] There's large number of HTTP metrics, and some Python ones too.  It would be good to have controls for limiting those in larger clusters, but I did not see any options for that in `prometheus_fastapi_instrumentator` API.

eero-t:
@Spycsh from you comment in the bug https://github.com/opea-project/GenAIComps/issues/1121#issuecomment-2581628392

I realized that changing the method on first metric access is racy.  It's possible that multiple threads end up in create method, before that method is changed to update one.  Meaning that multiple identical metrics would be created, and Prometheus would barf on that.

=> I'll add lock & check to handle that.

eero-t:
Rebased by `main`, no other changes.

@Spycsh Does this look fine / OK to merge, now that v1.2 got branched & tagged?

mkbhanda:
@eero-t does it make sense to have Guage for every service or just the end to end application, as in megaservice chatQnA?  Also are there not default settings for histogram buckets?

Spycsh (approved):

Spycsh:
LGTM.

eero-t:
> @eero-t does it make sense to have Guage for every service or just the end to end application, as in megaservice chatQnA?

Short answer:

There's no way to differentiate whether service instantiating orchestrator class is frontend, or middle service, so orchestrator metrics appear to whichever service instantiates its class.  This is the actual problem this PR tries to deal with (and is relevant for dashboards).

Long answer:

_Every_ OPEA service _does_ provide metrics, at least HTTP query stats, but those include also health and metric queries (regularly polled from the services by other k8s components), and do not include all relevant info for service SLAs.

Metrics this PR is concerned, are end-user request and token latencies, relevant for tracking service SLAs, as measured within the service orchestrator class (only place where they actually can be measured).

The problem that this PR deals with, is that service orchestrator is/was/can be used also OPEA components that do _not_ process tokens, so providing (zero valued) metrics about them would be misleading (especially if those services then show up unwanted in dashboards).  It also makes megaservice metric behaviour similar to (TEI/TGI/vLLM) inferencing services, which do not provide metrics until they've processed their first request.


> Was just curious why we did not have some default absolute value for histogram buckets for first token latency given it is like an SLA.

How well given buckets fit given service, depends completely on what kind of LLM model/params are used, and whether inferencing is accelerated, how many backend there are, and also to some extent how stressed the service is.

Prometheus already has defaults for histogram buckets, which are exponential.  They are good enough that you'll see metrics being spread to multiple buckets, regardless of used model/acceleration.

If one would want more details, buckets would need to be specified separately for each different service and its underlying HW configuration.  I.e. they would need to be externally configurable, and specified in some kind of service/HW profile.

 
> I did see some settings in GenAIEval (...) and get its really data collection first and only display determined in Grafana.

Values given to Prometheus quantile queries are _percentage_ thresholds, not the histogram (`le` = less than) bucket thresholds.  I.e. those dashboards should work regardless of what are the used metric value bucket thresholds.

How accurate the quantile information is depends both on how well the values are spread to buckets, and how well that matches quantile percentage thresholds though.

mkbhanda (approved):
LGTM and thank you @eero-t for the clarifications on Guage and default histogram buckets.

PR #1104, opened 2025-01-03
Add kubernetes deployment for GenAIComps

yongfengdu:
## Description

## Issues

## Type of change

## Dependencies

## Tests

### DIFF
--- a/comps/llms/deployment/kubernetes/README.md
+++ b/comps/llms/deployment/kubernetes/README.md
@@ -1,0 +1,11 @@
+# Deploy LLM microservice on Kubernetes cluster
+
+- You should have Helm (version >= 3.15) installed. Refer to the [Helm Installation Guide](https://helm.sh/docs/intro/install/) for more information.
+- For more deployment options, refer to [helm charts README](https://github.com/opea-project/GenAIInfra/tree/main/helm-charts#readme).
+
+## Deploy on Kubernetes
+
+```
+export HFTOKEN="insert-your-huggingface-token-here"
+helm install llm oci://ghcr.io/opea-project/charts/llm-uservice --set global.HUGGINGFACEHUB_API_TOKEN=${HFTOKEN} -f cpu-values.yaml
+```
--- a/comps/agent/deployment/kubernetes/gaudi-values.yaml
+++ b/comps/agent/deployment/kubernetes/gaudi-values.yaml
@@ -1,0 +1,38 @@
+# Copyright (C) 2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+# Accelerate inferencing in heaviest components to improve performance
+# by overriding their subchart values
+
+tgi:
+  enabled: true
+  accelDevice: "gaudi"
+  image:
+    repository: ghcr.io/huggingface/tgi-gaudi
+    tag: "2.0.6"
+  resources:
+    limits:
+      habana.ai/gaudi: 4
+  MAX_INPUT_LENGTH: "4096"
+  MAX_TOTAL_TOKENS: "8192"
+  CUDA_GRAPHS: ""
+  OMPI_MCA_btl_vader_single_copy_mechanism: "none"
+  PT_HPU_ENABLE_LAZY_COLLECTIVES: "true"
+  ENABLE_HPU_GRAPH: "true"
+  LIMIT_HPU_GRAPH: "true"
+  USE_FLASH_ATTENTION: "true"
+  FLASH_ATTENTION_RECOMPUTE: "true"
+  extraCmdArgs: ["--sharded","true","--num-shard","4"]
+  livenessProbe:
+    initialDelaySeconds: 5
+    periodSeconds: 5
+    timeoutSeconds: 1
+  readinessProbe:
+    initialDelaySeconds: 5
+    periodSeconds: 5
+    timeoutSeconds: 1
+  startupProbe:
+    initialDelaySeconds: 5
+    periodSeconds: 5
+    timeoutSeconds: 1
+    failureThreshold: 120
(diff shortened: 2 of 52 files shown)


comps/3rd_parties/tgi/deployment/kubernetes/README.md
```diff
@@ -0,0 +1,18 @@
+# Deploy tgi on kubernetes cluster
```
poussa, line 1: TGI in capital letters.

comps/llms/deployment/kubernetes/README.md
```diff
@@ -0,0 +1,11 @@
+# Deploy llm microservice on Kubernetes cluster
```
poussa, line 1: LLM in capital letters.

poussa (changes requested):

comps/llms/deployment/kubernetes/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+- You should have Helm (version >= 3.15) installed. Refer to the [Helm Installation Guide](https://helm.sh/docs/intro/install/) for more information.
+- For more deploy options, refer to [helm charts README](https://github.com/opea-project/GenAIInfra/tree/main/helm-charts#readme).
```
mkbhanda, line 4:
```suggestion
- For more deployment options, refer to [helm charts README](https://github.com/opea-project/GenAIInfra/tree/main/helm-charts#readme).
```

mkbhanda (commented):
Some questions ..

Any reason why no failureThreshold on the readinessProbe for CPU and gaudi values?

Why do we not mention any image for CPU .. is it in some common file. Likewise all the min/max token lengths.

What about model id with Gaudi?

Thank you @yongfengdu for this PR.

comps/3rd_parties/tgi/deployment/kubernetes/gaudi-values.yaml
```diff
@@ -0,0 +1,39 @@
...
+CUDA_GRAPHS: ""
+OMPI_MCA_btl_vader_single_copy_mechanism: "none"
+ENABLE_HPU_GRAPH: "true"
+LIMIT_HPU_GRAPH: "true"
+USE_FLASH_ATTENTION: "true"
+FLASH_ATTENTION_RECOMPUTE: "true"
+
+resources:
+  limits:
+    habana.ai/gaudi: 1
```
eero-t, line 25: This could also mention model, and corresponding requests for CPU & memory, to get better priority for the pod, and to avoid too much of CPU pods being put on Gaudi node (as taints are not used to keep non-device workloads out of it).

yongfengdu:
> Some questions ..
> 
> Any reason why no failureThreshold on the readinessProbe for CPU and gaudi values?
These values were tuned by @eero-t and I just borrowed them.
The default failureThreshold is 3.
For livenessProbe failure, the pod will be killed/restarted by kubelet, I think that's why the failureThreshold was set to a bigger value.
> 
> Why do we not mention any image for CPU .. is it in some common file. Likewise all the min/max token lengths.
There are default values in values.yaml, which is included in the helm chart. But it might be a good idea to mention them in cpu-values.yaml and gaudi-values.yaml, even if they are the same as default values.yaml.
I'll refine the *-values.yaml file.
> What about model id with Gaudi?
> 
> Thank you @yongfengdu for this PR.

eero-t:
>> Any reason why no failureThreshold on the readinessProbe for CPU and gaudi values?
>
> These values were tuned by @eero-t and I just borrowed them.

I was concentrating more on when things actually start working, rather than when they start failing, so it's a rather rough value.

> The default failureThreshold is 3.

They could be explicitly mentioned also for `readinessProbe`s.

For Gaudi, I think the readiness probe `failureThreshold` could be smaller, but I haven't tested it.

For CPU it's definitely better to keep it >1 because perf on CPU is so unpredictable (because underlying HW can differ and pods do not have fine-tuned resource requests / limits).

> For livenessProbe failure, the pod will be killed/restarted by kubelet, I think that's why the failureThreshold was set to a bigger value.

I've never seen OPEA pods deadlock, especially in a way that would be solved by restarting the pod.  I.e. liveness probe restarts just make things worse, as service may then never reach ready/live state => IMHO liveness probes are just harmful, and could be removed.

poussa:
Why do we have `kubernetes/deployment` in the pathname? It should be either `kubernetes` or `deployment`, not both. Or what is the logic here?

yongfengdu:
> Why do we have `kubernetes/deployment` in the pathname? It should be either `kubernetes` or `deployment`, not both. Or what is the logic here?

It's "deployment/kubernetes"
There is also a docker_compose directory under deployment(deployment/docker_compose), providing the instructions for docker environment.
See
https://github.com/opea-project/GenAIComps/tree/main/comps/3rd_parties/tgi/deployment

lianhao (commented):
for dataprep and retriever, the current value files will be obsolete in a couple of days.(new PR in infra repo will be submitted next week). Should we remove them from this PR?

yongfengdu:
I've verified the current *-values.yaml files works fine with latest released version (1.1), so I think it's reasonable to merge this change first, and do a follow up fix after helm charts changed.
We'll need to provide another fix anyway, unless we include all required changes after code restructure, which is impossible.
> for dataprep and retriever, the current value files will be obsolete in a couple of days.(new PR in infra repo will be submitted next week). Should we remove them from this PR?

lianhao (approved):
let's land-in this first to add the following CI tasks. Will update relevant values files later after GenAIInfra's helm chart is refactored for v1.2

chensuyue (approved):

chensuyue:
After the 0-nightly on push event ready, we can open the CI for this Repo.

PR #1122, opened 2025-01-07
add opentelemetry tracing into OPEA DAG and couple microservices code path related to ChatQnA

louie-tsai:
## Description

Added open telemetry tracing into OPEA DAG execute path and also couple microservices related to ChatQnA
The purpose is to show tracing timeline for each single request.
image

depend on https://github.com/opea-project/GenAIExamples/pull/1316

## Issues

 `n/a`.

## Type of change

## Dependencies

NA

## Tests

manually testing

### DIFF
--- a/comps/cores/mega/orchestrator.py
+++ b/comps/cores/mega/orchestrator.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import asyncio
+import contextlib
 import copy
 import json
 import os
@@ -14,7 +15,6 @@ import requests
 from fastapi.responses import StreamingResponse
 from prometheus_client import Gauge, Histogram
 from pydantic import BaseModel
-import contextlib
 
 from ..proto.docarray import LLMParams
 from ..telemetry.opea_telemetry import opea_telemetry, tracer
@@ -171,9 +171,13 @@ class ServiceOrchestrator(DAG):
 
     def wrap_iterable(self, iterable, is_first=True):
 
-        with tracer.start_as_current_span(f"llm_generate_stream"):
+        with tracer.start_as_current_span("llm_generate_stream"):
             while True:
-                with tracer.start_as_current_span(f"llm_generate_stream_first_token") if is_first else contextlib.nullcontext(): #  else tracer.start_as_current_span(f"llm_generate_stream_next_token")
+                with (
+                    tracer.start_as_current_span("llm_generate_stream_first_token")
+                    if is_first
+                    else contextlib.nullcontext()
+                ):  #  else tracer.start_as_current_span(f"llm_generate_stream_next_token")
                     try:
                         token = next(iterable)
                         yield token
@@ -282,9 +286,7 @@ class ServiceOrchestrator(DAG):
                 response = await session.post(endpoint, json=input_data)
             if response.content_type == "audio/wav":
                 audio_data = await response.read()
-                data = self.align_outputs(
-                    audio_data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs
-                )
+                data = self.align_outputs(audio_data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs)
             else:
                 # Parse as JSON
                 data = await response.json()
--- a/comps/cores/mega/orchestrator.py
+++ b/comps/cores/mega/orchestrator.py
@@ -14,9 +14,10 @@ import requests
 from fastapi.responses import StreamingResponse
 from prometheus_client import Gauge, Histogram
 from pydantic import BaseModel
+import contextlib
 
 from ..proto.docarray import LLMParams
-from ..telemetry.opea_telemetry import opea_telemetry
+from ..telemetry.opea_telemetry import opea_telemetry, tracer
 from .constants import ServiceType
 from .dag import DAG
 from .logger import CustomLogger
@@ -168,6 +169,21 @@ class ServiceOrchestrator(DAG):
             all_outputs.update(result_dict[prev_node])
         return all_outputs
 
+    def wrap_iterable(self, iterable, is_first=True):
+
+        with tracer.start_as_current_span(f"llm_generate_stream"):
+            while True:
+                with tracer.start_as_current_span(f"llm_generate_stream_first_token") if is_first else contextlib.nullcontext(): #  else tracer.start_as_current_span(f"llm_generate_stream_next_token")
+                    try:
+                        token = next(iterable)
+                        yield token
+                        is_first = False
+                    except StopIteration:
+                        # Exiting the iterable loop cleanly
+                        break
+                    except Exception as e:
+                        raise e
+
     @opea_telemetry
     async def execute(
         self,
@@ -211,14 +227,15 @@ class ServiceOrchestrator(DAG):
                 hitted_ends = [".", "?", "!", "。", "，", "！"]
                 downstream_endpoint = self.services[downstream[0]].endpoint_path
 
-            @opea_telemetry
             def generate():
                 token_start = req_start
                 if response:
                     # response.elapsed = time until first headers received
                     buffered_chunk_str = ""
                     is_first = True
-                    for chunk in response.iter_content(chunk_size=None):
+
+                    for chunk in self.wrap_iterable(response.iter_content(chunk_size=None)):
+
                         if chunk:
                             if downstream:
                                 chunk = chunk.decode("utf-8")
@@ -244,6 +261,7 @@ class ServiceOrchestrator(DAG):
                                 token_start = self.metrics.token_update(token_start, is_first)
                                 yield chunk
                             is_first = False
+
                     self.metrics.request_update(req_start)
                     self.metrics.pending_update(False)
 
@@ -260,19 +278,20 @@ class ServiceOrchestrator(DAG):
                 input_data = {k: v for k, v in input_data.items() if v is not None}
             else:
                 input_data = inputs
-            async with session.post(endpoint, json=input_data) as response:
-                if response.content_type == "audio/wav":
-                    audio_data = await response.read()
-                    data = self.align_outputs(
-                        audio_data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs
-                    )
-                else:
-                    # Parse as JSON
-                    data = await response.json()
-                    # post process
-                    data = self.align_outputs(data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs)
+            with tracer.start_as_current_span(f"{cur_node}_generate"):
+                response = await session.post(endpoint, json=input_data)
+            if response.content_type == "audio/wav":
+                audio_data = await response.read()
+                data = self.align_outputs(
+                    audio_data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs
+                )
+            else:
+                # Parse as JSON
+                data = await response.json()
+                # post process
+                data = self.align_outputs(data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs)
 
-                return data, cur_node
+            return data, cur_node
 
     def align_inputs(self, inputs, *args, **kwargs):
         """Override this method in megaservice definition."""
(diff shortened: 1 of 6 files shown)


comps/cores/mega/orchestrator.py
```diff
@@ -166,6 +168,7 @@ def process_outputs(self, prev_nodes: List, result_dict: Dict) -> Dict:
             all_outputs.update(result_dict[prev_node])
         return all_outputs
 
+    @opea_telemetry
```
Spycsh, line 171: in this `execute` method we basically has two cases:

1) streaming responses
2) non-streaming responses

for 2) , we can use this `opea_telemetry` decorator because the time span is selected correctly
for 1),  this `opea_telemetry` decorator may not work well because the Streaming response is directly returned (at that moment, the first token is not yielded at all!) before the first yield so the time span is not the actual execution time.

comps/cores/mega/orchestrator.py
```diff
@@ -208,6 +211,7 @@ async def execute(
                 hitted_ends = [".", "?", "!", "。", "，", "！"]
                 downstream_endpoint = self.services[downstream[0]].endpoint_path
 
+            @opea_telemetry
             def generate():
```
Spycsh, line 215: This is case 2) so simply add @opea_telemetry does not produce correct duration for `generate` here.

comps/cores/mega/orchestrator.py
```diff
@@ -301,6 +305,7 @@ def extract_chunk_str(self, chunk_str):
             chunk_str = chunk_str[: -len(suffix)]
         return chunk_str
 
+    @opea_telemetry
     def token_generator(self, sentence: str, token_start: float, is_first: bool, is_last: bool) -> str:
```
Spycsh, line 309: This is case 2). As shown from your Jaeger UI image, I'm afraid this is never traced?

comps/llms/src/text-generation/opea_llm_microservice.py
```diff
@@ -36,6 +37,7 @@
     host="0.0.0.0",
     port=9000,
 )
+@opea_telemetry
 @register_statistics(names=["opea_service@llm"])
 async def llm_generate(input: Union[LLMParamsDoc, ChatCompletionRequest, SearchedDoc]):
```
Spycsh, line 42: When `stream=True`. This is case 2) so the duration is not correct. Only correct when `stream=False`.

comps/llms/text-generation/vllm/langchain/llm.py
```diff
@@ -72,6 +72,7 @@ def post_process_text(text: str):
     host="0.0.0.0",
     port=9000,
 )
+@opea_telemetry
 async def llm_generate(input: Union[LLMParamsDoc, ChatCompletionRequest, SearchedDoc]):
```
Spycsh, line 76: When stream=True. This is case 2) so the duration is not correct. Only correct when stream=False.

comps/cores/mega/orchestrator.py
Spycsh, line 1:
```py
            async with session.post(endpoint, json=input_data) as response:
                if response.content_type == "audio/wav":
                    audio_data = await response.read()
                    data = self.align_outputs(
                        audio_data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs
                    )
                else:
                    # Parse as JSON
                    data = await response.json()
                    # post process
                    data = self.align_outputs(data, cur_node, inputs, runtime_graph, llm_parameters_dict, **kwargs)

```
 these are the lines that dorequest/reply from megaservice to microservice for non-streaming case 1). I think we should extract that lines to a function and add `@opea_telemetry`.

Spycsh:
Hi @louie-tsai the tracing implementation still have some problems:

1) streaming response tracing (LLM) shows incorrect duration
2) non-streaming generation (Embedding/Retriever) are not traced

Please check my comments above and feel free to tell me if there are anything not clear!

comps/cores/mega/orchestrator.py
eero-t: `self.metrics.request_update(req_start)` line in the code shows the place after which processing of **streaming** call ends.

eero-t:
IMHO review comments above are an indication that `ServiceOrchestrator` would benefit from refactoring, so that its working can be reliably measured / traced.  To me e.g. streaming and non-streaming cases being threaded differently feels like design bug.   Unfortunately I don't have any concrete idea how to fix that though. :-/

louie-tsai:
> Hi @louie-tsai the tracing implementation still have some problems:
> 
> 1. streaming response tracing (LLM) shows incorrect duration
> 2. non-streaming generation (Embedding/Retriever) are not traced
> 
> Please check my comments above and feel free to tell me if there are anything not clear!

@Spycsh 
For case 1, do you have suggestion how to enable opentelemetry tracing for streaming response tracing?
In the meantime, need your help to enable tracing for non-streaming generation parts.

comps/cores/mega/orchestrator.py
louie-tsai: let me try to have a draft for your proposal

louie-tsai: @Spycsh  one quick question, if we put @open_telemetry before execute. why non-streaming codes are the traced?

Spycsh: @open_telemetry on a non-streaming function will trace the time span between the function called/returned and it works correctly because the moment of return is correct.

louie-tsai (commented):
@Spycsh  Thanks for your changes. It looks good to me. Let me know if the codes are ready to test.

louie-tsai:
> @louie-tsai I've draft/fix the correct streaming/non-streaming tracing implementation. Here are some screenshots:
> 
> 1. non-streaming
> 
> image
> 
> 2. streaming
> 
> image
> 
> image
> 
> I will get back to you about some details/tradeoffs later.

thanks.  it looks good.  especially we have node name in the span. just a quick question, for llm duration in non-streaming case, does it include all tokens since the duration is relative big?

eero-t:
> thanks. it looks good. especially we have node name in the span. just a quick question, for llm duration in non-streaming case, does it include all tokens since the duration is relative big?

Yes, it's the complete request reply.  Stream option for the query means that reply should return each token separately.

Spycsh:
> > @louie-tsai I've draft/fix the correct streaming/non-streaming tracing implementation. Here are some screenshots:
> > 
> > 1. non-streaming
> > 
> > image
> > 
> > 2. streaming
> > 
> > image
> > image
> > I will get back to you about some details/tradeoffs later.
> 
> thanks. it looks good. especially we have node name in the span. just a quick question, for llm duration in non-streaming case, does it include all tokens since the duration is relative big?

Yes it does. And also if you look at the first token latency in the second screenshot in streaming case, namely the third screenshot I give, it displays 500ms total latency, 92ms first token latency. The non-streaming screenshot in the first screenshot is also around 537ms, <10% deviation from streaming case. I thought it is just acceptable error affected by the warmup. Also the same reason to TEI embedding service latency.

Spycsh:
> @Spycsh Thanks for your changes. It looks good to me. Let me know if the codes are ready to test.

I think you should firstly fix the conflict and trigger the CI. The code now I thought is relatively complete but you can of course customize to your need if you have better ideas!

louie-tsai:
> > @Spycsh Thanks for your changes. It looks good to me. Let me know if the codes are ready to test.
> 
> I think you should firstly fix the conflict and trigger the CI. The code now I thought is relatively complete but you can of course customize to your need if you have better ideas!

thanks. fixed the conflict and ran it on Gaudi3 by patching old codes.

I saw two seperate spans. one is handle_request and the other is llm_generation_stream
image

there are embedding and retriever in handle_request. it looks good. thanks for the changes.
image
also saw first_token in llm span.
image

I will generate the docker images based on the latest codes and run on Gaudi3 later.

louie-tsai:
@chensuyue  those two test failures don't seem to be related to this PR. are there any known issue related to this issues?

eero-t:
> @chensuyue those two test failures don't seem to be related to this PR. are there any known issue related to this issues?

They are not related to changes in this PR.  Lot of PRs get pushed in just before release freeze, so there's large number of changes needing stabilization...

Annotations check issue is some bug in CI test `sed` usage itself.

"llms_text-generation_native_llamaindex_on_intel_hpu" CI test output:
```
#11 17.39 ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
#11 17.39 neural-compressor-pt 3.1 requires pydantic==1.10.13, but you have pydantic 2.10.5 which is incompatible.
...
[ llm-native ] HTTP status is not 200. Received status was 500
...
 [WARNING|utils.py:212] 2025-01-14 10:19:23,574 >> optimum-habana v1.15.0 has been validated for SynapseAI v1.19.0 but habana-frameworks v1.18.0.524 was found, this could lead to undefined behavior!
[WARNING|utils.py:229] 2025-01-14 10:19:23,627 >> Could not run `hl-smi`, please follow the installation guide: https://docs.habana.ai/en/latest/Installation_Guide/index.html.
...
   File "/home/user/.local/lib/python3.10/site-packages/huggingface_hub/file_download.py", line 452, in http_get
    for chunk in r.iter_content(chunk_size=constants.DOWNLOAD_CHUNK_SIZE):
  File "/usr/local/lib/python3.10/dist-packages/requests/models.py", line 822, in generate
    raise ChunkedEncodingError(e)
requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(2509955832 bytes read, 1354770576 more expected)', IncompleteRead(2509955832 bytes read, 1354770576 more expected))
+ exit 1
Error: Process completed with exit code 1.
```

shows several issues:
* Version mismatches between dependencies of Python modules
* Version mismatches between Habana host and container driver versions
* File download from HF being interrupted

Last one is the main issue I think, but it could be temporary, e.g. due to CI exceeding HF rate limit...

Spycsh (approved):

louie-tsai:
also add a span for llm_asyn_generate to make a connection to llm_generate_stream
Screenshot from 2025-01-15 17-38-50

chensuyue:
I am checking the CI issue, this seems to be a scripts issue, but only happens on this PR. 
image

chensuyue:
I found the root cause, while running the CI, we use `comps_path=$(python -c 'import comps; print(comps.__path__[0])')` to get the comps path. But in your PR, the changes cause the special return, image
 which comes from https://github.com/opea-project/GenAIComps/blob/ac650c57237fca18b4f0ef5a38ddf826cf543d0c/comps/cores/telemetry/opea_telemetry.py#L46 .  
Those return are not correct part of the comps_path, 
`find . -name "test*.py" | sed "s,\.\/,python -m pytest --cov=\"${comps_path}\" --cov-report term --cov-report xml:coverage.xml --cov-append -vs --disable-warnings ,g" > run.sh`
so it reported, 
image.

chensuyue:
I don't recommend to print those logs while import comps. 
https://github.com/opea-project/GenAIComps/blob/ac650c57237fca18b4f0ef5a38ddf826cf543d0c/comps/cores/telemetry/opea_telemetry.py#L46

Spycsh:
> I found the root cause, while running the CI, we use `comps_path=$(python -c 'import comps; print(comps.__path__[0])')` to get the comps path. But in your PR, the changes cause the special return, image which comes from
> 
> https://github.com/opea-project/GenAIComps/blob/ac650c57237fca18b4f0ef5a38ddf826cf543d0c/comps/cores/telemetry/opea_telemetry.py#L46
> 
> .
> Those return are not correct part of the comps_path,
> `find . -name "test*.py" | sed "s,\.\/,python -m pytest --cov=\"${comps_path}\" --cov-report term --cov-report xml:coverage.xml --cov-append -vs --disable-warnings ,g" > run.sh`
> so it reported,
> image.

Thanks @chensuyue . @louie-tsai I think we can simply remove that printing statement. It was very early code for debugging telemetry functionalities and are no longer needed for current use.

louie-tsai:
@chensuyue  fixed the issue according to your finding.
however, still a strange issue there.
`
Error response from daemon: No such container: test-comps-retrievers-neo4j-llama-index-tgi
`

any idea?

yinghu5 (approved):
Looks goods, thanks

PR #1127, opened 2025-01-08
Add Dockerfile for comps-base image

eero-t:
## Description

Add Docker file for a base image that can be used by GenAIExamples applications.

With a shared base image, CI can do test builds much faster, they take much less space locally and in registry, and images sharing the same base download much faster.

With this, almost all of the 20 application Dockerfiles, e.g. this one:
https://github.com/opea-project/GenAIExamples/blob/main/ChatQnA/Dockerfile

Can be reduced just to:
```
# Copyright (C) 2025 Intel Corporation

ARG BASE_TAG=latest
FROM opea/comps-base:$BASE_TAG

COPY ./chatqna.py $HOME/chatqna.py
ENTRYPOINT ["python", "chatqna.py"]
```

(For releases, `docker build` command can be given e.g. `--build-arg=BASE_TAG=1.2` option.)

## Issues

This is first part in fixing image sizes and how much time and disk space is used for them.  There are multiple tickets about the image sizes.

## Type of change

## Dependencies

`n/a`

## Tests

Tested manually that the added Dockerfile, and ChatQnA & DocSum Dockerfiles relying on it build, and that the resulting ChatQnA & DocSum applications work fine.

## Next steps

Somebody (else) needs to:
* Add checking & building of this image to CI
* Do nightly builds of it and upload them both to CI and DockerHub registries for applications
* Do similar simplification / optimization for rest of the Dockerfiles (ones in this repo, and UI ones in GenAIExamples)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+# Base image for GenAIComps based OPEA Python applications
+# Build: docker build -t opea/comps-base -f Dockerfile .
+
+ARG IMAGE_NAME=python
+ARG IMAGE_TAG=3.11-slim
+
+FROM ${IMAGE_NAME}:${IMAGE_TAG} AS base
+
+ENV HOME=/home/user
+
+RUN useradd -m -s /bin/bash user && \
+    mkdir -p $HOME && \
+    chown -R user $HOME
+
+# get security updates
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get clean autoclean && \
+    apt-get autoremove -y && \
+    rm -rf /var/lib/apt/lists/*
+
+WORKDIR $HOME
+
+COPY *.toml *.py *.txt *.md LICENSE ./
+
+RUN pip install --no-cache-dir --upgrade pip setuptools && \
+    pip install --no-cache-dir -r requirements.txt
+
+COPY comps/ comps/
+
+ENV PYTHONPATH=$PYTHONPATH:$HOME
+
+USER user
+
+ENTRYPOINT ["sh", "-c", "set && ls -la"]


eero-t:
With current `opea/chatqna` Dockerfile, the image size is 828MB.

The image size resulting from this new base image + the ChatQnA Dockerfile example, is 436MB.

=> image size halves, as it drops installation of redundant package (Mesa 3D libs, Git and their deps, e.g. Perl).

Every additional application image will take only fraction of that, because base layer already exists.  Docker / crictl still will list the combines size though (`docker` lists uncompressed one, `crictl` compressed one).

eero-t:
> => image size halves, as it drops installation of redundant package (Mesa 3D libs, Git and their deps, e.g. Perl).

Earlier https://github.com/opea-project/GenAIExamples/pull/1031 PR showed that CI tests for all apps pass even without these extra packages.  That's why this new base image does not include them.

eero-t:
Several Dockerfiles in "GenAIExamples" include also following which I've dropped from the base:
```
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
...
RUN echo 'ulimit -S -n 999999' >> ~/.bashrc
```

Because `hadolint` does not complain about former missing, and because all these containers run Python directly instead of Bash.   But having them there should not harm either, as they take no space, unlike the unused packages.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM ${IMAGE_NAME}:${IMAGE_TAG} AS base
+
+ENV HOME=/home/user
+
+RUN useradd -m -s /bin/bash user && \
+    mkdir -p $HOME && \
+    chown -R user $HOME
+
+# get security updates
+RUN apt-get update && apt-get upgrade -y && \
```
ashahba, line 19: Once the `upgrade` is done, some old packages may become stale and no longer needed and it that case:
```
apt-get autoclean && \
apt-get autoremove -y
```
would cleanup even furthur.

ashahba (changes requested):
Thanks @eero-t for this PR.
My feedback is only a minor one.

Dockerfile
eero-t: I don't think that is necessary in this particular case.  `python:<major>.<minor>-slim` images are based on Debian _stable_, the Python itself comes from some binary tarball, so only Debian base packages are in the  package database.  Base packages are mostly marked as `Essential`, and mostly lack dependencies.   They could disappear only on dist-upgrade, not on plain upgrade.

But I can add it if you feel strongly about it, or if you think we may use some more complicated base in future?

poussa:
This is good and should be merged ASAP. Only then examples can start using it.

Dockerfile
eero-t: But on the other, it does no harm either => added.

(And somebody could copy-pastes the update command to a container where it can help...)

eero-t:
Rebased to `main`.

eero-t:
> This is good and should be merged ASAP. Only then examples can start using it.

Next steps would be:
* adding this to nightly build images,
* pushing the new image to registry used by CI (in this and GenAIExamples repo), and
* pushing that image also to DockerHub OPEA project.

However, I have no clue how to achieve that, so unfortunately somebody else needs to do that.

Only after that, CI can pass for https://github.com/opea-project/GenAIExamples/pull/1369 so that it can also be merged.

fidencio:
While merged is block, @eero-t, what about squashing everything into the same commit? Just a suggestion, not a requirement.

eero-t:
> While merged is block, @eero-t, what about squashing everything into the same commit? Just a suggestion, not a requirement.

Done.

ashahba (approved):
LGTM!

fidencio (approved):
lgtm, thanks @eero-t!

chensuyue (commented):
This design LGTM, just need more time to update CICD.

eero-t:
Rebased to `main`.   Several weeks have passed, and 1.2 tagged.  Wouldn't it be high time to merge this?

mkbhanda (approved):
LGTM

eero-t:
Next steps would be:
* adding the merged container image spec to be built with other nightly images,
* pushing the resulting image to registry used by CI (in this and GenAIExamples repo), and
* pushing it also to DockerHub OPEA project.

(However, I have no clue how to achieve that, so unfortunately somebody else needs to do that.)

PR #1133, opened 2025-01-10
Fix TEI embedding docker compose health check issue

lvliang-intel:
## Description

Fix TEI embedding docker compose health check issue

## Issues

`n/a`

## Type of change

## Dependencies

None

## Tests

Local test and CI test


ashahba (approved):
LGTM!

fidencio:
Would it be possible to add the description of what's being fixed and how it's being fixed as part of the commit message?

By doing so, we're helping our future selves to solve similar issues, and also to understand better why a change was made just by navigating through the project's history.

rbrugaro:
@fidencio There were two issues: first, the TEI Docker image doesn't include the curl command, causing the health check to fail; second, the health check requires using host_ip instead of localhost. Since the CI test script uses docker run instead of the docker-compose.yaml, these issues were not identified earlier. This PR solves both

rbrugaro:
Needed for [PR #16666](https://github.com/run-llama/llama_index/pull/16666)

letonghan (approved):

lkk12014402 (approved):

PR #1134, opened 2025-01-10
MultimodalQnA image query, pdf, and dynamic ports

mhbuehler:
## Description

According to the [RFC](https://github.com/opea-project/docs/blob/main/community/rfcs/24-10-02-GenAIExamples-001-Image_and_Audio_Support_in_MultimodalQnA.md)'s Phase 2 plan, this PR adds image query support, PDF ingestion support, and dynamic ports to the microservices used by MultimodalQnA. This PR goes with this one in [GenAIExamples](https://github.com/opea-project/GenAIExamples/pull/1381).

## Issues

[RFC](https://github.com/opea-project/docs/blob/main/community/rfcs/24-10-02-GenAIExamples-001-Image_and_Audio_Support_in_MultimodalQnA.md)

## Type of change

## Dependencies

[pymupdf](https://github.com/pymupdf/PyMuPDF) is new for the dataprep microservice, but it's not new to GenAIComps.

## Tests

Tests were added to the following scripts:

- tests/dataprep/test_dataprep_multimodal_redis_langchain.sh
- tests/embeddings/test_embeddings_multimodal.sh
- tests/lvms/test_lvms_llava.sh
- tests/lvms/test_lvms_tgi-llava_on_intel_hpu.sh
- tests/retrievers/test_retrievers_multimodal_redis_langchain.sh
- tests/retrievers/test_retrievers_redis.sh


comps/dataprep/multimodal/redis/langchain/prepare_videodoc_redis.py
```diff
@@ -301,7 +304,53 @@ def prepare_data_and_metadata_from_annotation(
...
+    3. Instead of time of frame in ms, we return the PDF page index through
+       the pre-existing time_of_frame_ms metadata key to maintain compatibility.
+    """
+    text_list = []
+    image_list = []
+    metadatas = []
+    for frame in annotation:
+        page_index = frame["frame_no"]
+        image_index = frame["sub_video_id"]
+        path_to_frame = os.path.join(path_to_frames, f"page{page_index}_image{image_index}.png")
```
tileintel, line 323: @mhbuehler Anyway, this is a new function. Why do we need to preserve the name of local variables relating to frames/videos (e.g., path_to_frame, video_id)? This might cause confusion.

comps/asr/src/integrations/dependency/whisper/whisper_server.py
```diff
@@ -106,7 +106,7 @@ async def audio_transcriptions(
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--host", type=str, default="0.0.0.0")
-    parser.add_argument("--port", type=int, default=7066)
+    parser.add_argument("--port", type=int, default=os.getenv("WHISPER_PORT", 7066))
```
Spycsh, line 109: The internal container server port can be forwarded to any arbitrary host port. Why we need this also to be configurable?

comps/asr/src/opea_asr_microservice.py
```diff
@@ -34,7 +34,7 @@
     service_type=ServiceType.ASR,
     endpoint="/v1/audio/transcriptions",
     host="0.0.0.0",
-    port=9099,
+    port=int(os.getenv("ASR_PORT", 9099)),
     input_datatype=Base64ByteStrDoc,
```
Spycsh, line 38: Same question as above.

comps/dataprep/multimodal/redis/langchain/prepare_videodoc_redis.py
mhbuehler: We don't, I'll fix this. Thanks!

comps/asr/src/integrations/dependency/whisper/whisper_server.py
mhbuehler: I see your point. Our intention with `WHISPER_PORT` and `ASR_PORT` was to enable full customization to the user, but as you pointed out, hard-coding the internal server ports doesn't really matter because they can be forwarded to different external ports.

The change to `ASR_PORT` is now irrelevant because it's been removed from GenAIExamples, so we will revert that change. Let me know if you prefer to have us revert `WHISPER_PORT` configurability and we'll do the same for it.

comps/asr/src/opea_asr_microservice.py
mhbuehler: We will revert this.

comps/asr/src/integrations/dependency/whisper/whisper_server.py
Spycsh: Hi @mhbuehler , I think this should also be reverted, to keep redundant env variables as few as possible. Also I think even if users want to run the python script directly, they should not use an env variable like WHISPER_PORT but CAN use `python whisper_server.py --port XXX` for port customization.

dmsuehir: @Spycsh  Thanks, that makes sense. I pushed an update to revert this change.

dmsuehir (approved):

chensuyue (approved):

lkk12014402 (approved):

PR #1193, opened 2025-01-21
Bug fix neo4j dataprep 

rbrugaro:
Fix stored value for summary into graph DB to store actual text instead of coroutine obj

Added missing await statement to llm.achat call.

- [x ] Bug fix (non-breaking change which fixes an issue)


letonghan (approved):

ashahba (approved):
LGTM!

PR #1256, opened 2025-02-04
fix tei embedding and tei reranking bug

minmin-intel:
## Description

TEI embedding and reranking microservices throw error: [   ERROR] - opea_embedding_microservice - Error during embedding invocation: Task text-embedding has no recommended model. Please specify a model explicitly. This is due to model (ie endpoint url) was not passed in the post method.

## Issues

https://github.com/opea-project/GenAIComps/issues/1255

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would break existing design and interface)
- [ ] Others (enhancement, documentation, validation, etc.)

## Dependencies

No new dependencies

## Tests

UTs

PR #1271, opened 2025-02-07
Fix Dataprep Ingest Data Issue.

letonghan:
## Description

Fix Dataprep Ingest Data Issue.

Root Cause:
The package of `langchain_huggingface` updated, caused different output of `HuggingFaceEndpointEmbeddings.embed_documents`.

Trace:
1. The update of `langchain_huggingface.HuggingFaceEndpointEmbeddings` caused the wrong size of embedding vectors.
2. Wrong size vectors are wrongly saved into Redis database, and the indices are not created correctly.
3. The retriever can not retrieve data from Redis using index due to the reasons above.
4. Then the RAG seems `not work`, for the file uploaded can not be retrieved from database.

Solution:
Replace all of the `langchain_huggingface.HuggingFaceEndpointEmbeddings` to `langchain_community.embeddings.HuggingFaceInferenceAPIEmbeddings`, and modify related READMEs and scirpts.

## Issues

https://github.com/opea-project/GenAIExamples/issues/1473
https://github.com/opea-project/GenAIExamples/issues/1482

## Type of change

## Dependencies

None

## Tests

Local tested


comps/dataprep/src/integrations/redis.py
```diff
@@ -187,8 +188,15 @@ def ingest_chunks_to_redis(file_name: str, chunks: List):
...
     if TEI_EMBEDDING_ENDPOINT:
+        if not HUGGINGFACEHUB_API_TOKEN or not EMBED_MODEL:
+            raise HTTPException(
+                status_code=400,
+                detail="You MUST offer the `HUGGINGFACEHUB_API_TOKEN` and the `EMBED_MODEL` when using `TEI_EMBEDDING_ENDPOINT`.",
+            )
         # create embeddings using TEI endpoint service
-        embedder = HuggingFaceEndpointEmbeddings(model=TEI_EMBEDDING_ENDPOINT)
+        embedder = HuggingFaceInferenceAPIEmbeddings(
+            api_key=HUGGINGFACEHUB_API_TOKEN, model_name=EMBED_MODEL, api_url=TEI_EMBEDDING_ENDPOINT
```
lianhao, line 198: Ask the user to set `embed_model` in case of TEI_EMBEDDING_ENDPOINT would bring additional end user burden and/or prone to error.

We can get the model_name information in the runtime from a simple `curl http://<tei_embedding_point>/info`.

https://github.com/opea-project/GenAIComps/pull/1256 could be another way to avoid setting `embed_model`

lianhao (changes requested):

comps/dataprep/src/integrations/redis.py
letonghan: Thanks, good point. 
Will modify to get `EMBED_MODEL` automatically from `TEI_EMBEDDING_ENDPOINT`.

lianhao:
One more thing, I noticed that retriever is also using the `HuggingFaceEndpointEmbeddings`, should we change the retriever to use `HuggingFaceInferenceAPIEmbeddings` if the latest `HuggingFaceEndpointEmbeddings` is buggy?

letonghan:
> One more thing, I noticed that retriever is also using the `HuggingFaceEndpointEmbeddings`, should we change the retriever to use `HuggingFaceInferenceAPIEmbeddings` if the latest `HuggingFaceEndpointEmbeddings` is buggy?

The `embed_document` function is not used in Retriever component, so it does not affect retriever's functionality.

lvliang-intel (approved):

lianhao (approved):

PR #1280, opened 2025-02-12
Fix metric id issue when initializing multiple Orchestrator instances

Spycsh:
## Description

- Fix the wrong _instance_id handling in https://github.com/opea-project/GenAIComps/pull/1092
- essential for https://github.com/opea-project/GenAIExamples/pull/1528 UT pass

## Issues

as above
## Type of change

## Dependencies

na
## Tests

ut

### DIFF
--- a/comps/cores/mega/orchestrator.py
+++ b/comps/cores/mega/orchestrator.py
@@ -28,15 +28,13 @@ LOGFLAG = os.getenv("LOGFLAG", False)
 
 
 class OrchestratorMetrics:
-    # Need an instance ID for metric prefix because:
-    # - Orchestror instances are not named
-    # - CI creates several orchestrator instances
+    # Need an static class-level ID for metric prefix because:
     # - Prometheus requires metrics (their names) to be unique
     _instance_id = 0
 
     def __init__(self) -> None:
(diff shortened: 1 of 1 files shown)


WenjiaoYue (approved):

letonghan (approved):

comps/cores/mega/orchestrator.py
```diff
@@ -28,15 +28,13 @@
 
 
 class OrchestratorMetrics:
-    # Need an instance ID for metric prefix because:
-    # - Orchestror instances are not named
-    # - CI creates several orchestrator instances
```
eero-t, line 33: These are the reasons why this bug needed fixing, why they are removed?

(Only CI creates multiple Orchectrator, and therefore also multiple OrchestratorMetrics instances.)

Spycsh: No, not only CI, check DocSum. https://github.com/opea-project/GenAIExamples/blob/main/DocSum/docsum.py#L137

Also even we do not have it in production, we should also remove the comment that contain "only CI needs" because initializing multiple Orchestrator instances is allowed not only for CI but for flexible usage in production.

PR #1290, opened 2025-02-14
Fix telemetry connection issue when disabling telemetry

Spycsh:
## Description

- use ENABLE_OPEA_TELEMETRY to control whether to enable open telemetry, default false
- fix the issue that logs always show telemetry connection error with each request when telemetry is disabled
- ban the above error propagation to microservices when telemetry is disabled

## Issues

https://github.com/opea-project/GenAIExamples/issues/1552

## Type of change

## Dependencies

na
## Tests

ut

### DIFF
--- a/comps/cores/telemetry/opea_telemetry.py
+++ b/comps/cores/telemetry/opea_telemetry.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2024 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
+import contextlib
 import inspect
 import os
 from functools import wraps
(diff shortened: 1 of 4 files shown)


comps/cores/telemetry/opea_telemetry.py
```diff
@@ -13,6 +14,8 @@
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
 from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
 
+ENABLE_OPEA_TELEMETRY = os.getenv("ENABLE_OPEA_TELEMETRY", "false").lower() == "true"
```
lianhao, line 17: Instead of using 2 env variable `ENABLE_OPEA_TELEMETRY`, `TELEMETRY_ENDPOINT` to control the tracing, can we use just single env var `TELEMETRY_ENDPOINT` to control the tracing. When `TELEMETRY_ENDPOINT` is empty/None, it means the tracing is OFF. And we should make the tracing by default OFF.

Spycsh: I prefer not to because for one thing, the example side already use that ENABLE_OPEA_TELEMETRY flag
 to control whether to use telemetry, and I'm not sure whether removing it will cause some complaints from users that already use the telemetry. For another, I think a bit redundancy can help future contributors be easily aware of one fact that telemetry can be on/off with this flag. Otherwise we have to explain this rule in README or somewhere.

lvliang-intel (approved):

ZePan110 (approved):

letonghan (approved):

lkk12014402 (approved):

Issue #1287, opened 2025-02-13
[Bug] should NOT use synchronized I/O in GenAIComps microservices


lianhao:
### Priority

P2-High

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Installation method

- [ ] Pull docker images from hub.docker.com
- [ ] Build docker images from source
- [x] Other

### Deploy method

- [ ] Docker
- [ ] Docker Compose
- [ ] Kubernetes Helm Charts
- [x] Other

### Running nodes

Single Node

### What's the version?

git commit: 23b2be2

### Description

I've just happened to notice that the GenAIComp's core micro_service is implemented by python3's `asyncio.` However, I still see in many GenAIComp's microservices implementations, synchronous IO functions, e.g. `time.sleep`, `requests.post` etc. are being used. Using synchronous IO functions in `asyncio` would block the event loop for its duration, resulting the other coroutines not runnable. We should replace those synchronous IO with their asynchronous counterpart, e.g. `asyncio.sleep`, `aiohttp`, etc.

### Reproduce steps

n/a

### Raw log

```shell

```

lvliang-intel:
@XinyaoWa,
please refine some microservices according to lianhao's suggestion.

lianhao:
@XinyaoWa doing a 'grep -r requests.post' shows there are still bunch of synchronized calls. Could you please verify those synchronized calls are legit.
```
grep -r 'requests.post' comps/
comps/asr/src/integrations/whisper.py:            requests.post, f"{self.base_url}/v1/audio/transcriptions", files=files, data=data
comps/asr/src/integrations/dependency/whisper/check_whisper_server.py:response = requests.post(url=endpoint, data=json.dumps(inputs), proxies={"http": None})
comps/asr/src/check_asr_server.py:        response = requests.post(endpoint, headers=headers, data=data, files=files)
comps/animation/src/check_animation_server.py:response = requests.post(url=endpoint, json=data, headers={"Content-Type": "application/json"}, proxies={"http": None})
comps/third_parties/wav2lip/src/check_wav2lip_server.py:response = requests.post(url=endpoint, data=json.dumps(inputs), proxies={"http": None})
comps/lvms/src/integrations/video_llama.py:        response = requests.post(url=f"{self.base_url}/generate", params=params, proxies={"http": None}, stream=True)
[...]

PR #1300, opened 2025-02-18
Refine synchronized I/O in asynchronous functions

XinyaoWa:
## Description

- Update time.sleep and requests.post to asynchronous mode
- Keep unchanged if they aren't in asynchronous function links

## Issues

https://github.com/opea-project/GenAIComps/issues/1287

## Type of change

## Dependencies

## Tests


comps/asr/src/integrations/whisper.py
```diff
@@ -57,7 +58,9 @@ async def invoke(
         }
 
         # Send the file and model to the server
-        response = requests.post(f"{self.base_url}/v1/audio/transcriptions", files=files, data=data)
+        response = await asyncio.to_thread(
```
lianhao, line 61: Just out of curiosity. Why all others are replacing `requests` with `aiohttp`, except here?

comps/lvms/src/integrations/llava.py
```diff
@@ -112,11 +113,13 @@ async def invoke(
 
         inputs = {"img_b64_str": img_b64_str, "prompt": prompt, "max_new_tokens": max_new_tokens}
         # forward to the LLaVA server
-        response = requests.post(url=f"{self.base_url}/generate", data=json.dumps(inputs), proxies={"http": None})
+        async with aiohttp.ClientSession() as session:
```
Spycsh, line 116: Think it will be more efficient if we initialize this session in __init__ and here we edit it as `async with self.session.post...`!

comps/asr/src/integrations/whisper.py
XinyaoWa: Firstly, both aiohttp and asyncio.to_thread can work, I also find they both exist in original codes. Secondly, there exists some API and usage difference between requests.post and aiohttp. Thus in some cases, we just use asyncio.to_thread to avoid too much alignment and debug effort.

comps/cores/mega/orchestrator.py
```diff
@@ -259,51 +259,51 @@ async def execute(
...
-                    stream=True,
-                    timeout=1000,
-                )
+                async with aiohttp.ClientSession() as session:
+                    response = await session.post(
+                        url=endpoint,
+                        data=json.dumps(inputs),
+                        headers={"Content-type": "application/json"},
+                        proxy=None,
+                        timeout=aiohttp.ClientTimeout(total=1000),
```
mkbhanda, line 268: is there danger of this timeout 1000 being inadequate based on input length?

comps/dataprep/src/integrations/elasticsearch.py
```diff
@@ -142,7 +142,7 @@ def search_by_filename(self, file_name: str) -> bool:
 
         return results["hits"]["total"]["value"] > 0
 
-    def ingest_doc_to_elastic(self, doc_path: DocPath) -> None:
+    async def ingest_doc_to_elastic(self, doc_path: DocPath) -> None:
```
mkbhanda, line 145: General question, if we are given a list of documents, can we process/injest them in parallel? Further, could one split a document into chunks and process the chunks in parallel? Would certainly need load balancing over an embedding service and serialize at the point of inserting in the vector database and use write locks etc.

comps/dataprep/src/utils.py
```diff
@@ -319,47 +320,50 @@ def load_csv(input_path):
...
     png_path = svg_path.replace(".svg", ".png")
     cairosvg.svg2png(url=svg_path, write_to=png_path)
-    text = load_image(png_path)
+    text = await load_image(png_path)
     os.remove(png_path)
     return text
 
 
-def document_loader(doc_path):
+async def document_loader(doc_path):
```
mkbhanda, line 352: why no await for load_pdf, load_html etc?

comps/lvms/src/integrations/llava.py
lianhao: @Spycsh From Kubernetes(K8s) perspective, I don't think make the session persistent is a good idea. I assume a persistent session will result in a persistent TCP connection. If so,  it will prevent the out-of-box K8s to do the traffic load-balancing if we have multiple same inference pods for this lvm_llava service to talk to. All the requests will be routed to a single inference pod, and other pods will sit idle all the time. Unless you add some client-side load-balancing logic in your logic here, or we have some k8s cluster admin to deploy/configure service meshes.

I would suggest to keep shot lived sessions to make leverage the K8s's out-of-box load balancing, unless we're sure the TCP connection time is the main reason of performance bottle neck.

Spycsh: @lianhao if that really sticks k8s load-balancing yes we can keep the short-lived session as the code is! Better if we can  further look into the "stick" issue and if that is the case, we can have a flag to control whether to use diff approaches and leave room to reusing TCP connection on single node for optimized perf.

comps/dataprep/src/utils.py
XinyaoWa: Seems these two functions don't contain synchronized time-cost operation such as "requests.post", thus don't need to add.

comps/dataprep/src/integrations/elasticsearch.py
XinyaoWa: Good idea. This may need owner @letonghan  for an suggestion~

lianhao (approved):

lvliang-intel (approved):

eero-t:
This PR breaks (at least) ChatQnA and Docsum, they return empty result after this.

Reverting `orchestrator.py` to a commit just before this was merged, restores their functionality.

eero-t:
> This PR breaks (at least) ChatQnA and Docsum, they return empty result after this.

Maybe verifying that streaming response included all the expected tokens, would catch this issue in CI?
```
--- a/tests/cores/mega/test_service_orchestrator_streaming.py
+++ b/tests/cores/mega/test_service_orchestrator_streaming.py
@@ -64,6 +64,7 @@ class TestServiceOrchestratorStreaming(unittest.IsolatedAsyncioTestCase):
         async for k in response.__reduce__()[2]["body_iterator"]:
             self.assertEqual(self.service_builder.extract_chunk_str(k).strip(), res_expected[idx])
             idx += 1
+        self.assertEqual(i, len(res_expected))
 
     def test_extract_chunk_str(self):
         res = self.service_builder.extract_chunk_str("data: [DONE]\n\n")
```

Issue #1367, opened 2025-03-05
[Feature] Update Agent memory integrations with refactored OPEA DataStore component when ready


rbrugaro:
### Priority

Undecided

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Running nodes

Single Node

### Description

Feedback management, prompt registry and chat history components are going through a consolidation and Data Store refactor. 
I noticed [Agent/integrations/storage](https://github.com/opea-project/GenAIComps/blob/main/comps/agent/src/integrations/storage/persistence_redis.py) has a significant overlap. We should replace storage in agent component to use OPEA components and make it simple that agent leverage OPEA data store microservice rather than redefine them locally. 

cc: @lkk12014402 @ftian1 @minmin-intel @aMahanna @xiguiw

ftian1:
@rbrugaro could you pls mark the priority? shall we target it to v1.3?

yinghu5:
@rbrugaro @ftian1 thank you for raising the fature.  Just remind for any update

joshuayao:
It's not the priority for now, closed.

PR #1493, opened 2025-03-31
feature: OpeaStore Class

aMahanna:
## Description

Draft PR to introduce a Client API for basic CRUD functionality around databases.

Feedback welcomed.

```python
from comps.cores.storage import opea_store

store = opea_store(name="arangodb") # "redis", "mongodb", etc.

result = store.save_document({"foo": "bar"})

store.save_documents([{"foo": "baz"}, {"foo": "faz"}])

store.update_document({"_id": result["_id"], "foo": "bar!!!"})

store.update_documents(...)

result_2 = store.get_document_by_id(result["_id"])

assert result != result_2 # "foo" attribute now has different value

store.get_documents_by_ids([...])

result_3 = store.search(key="foo", value="bar!!!")[0]

assert result_2 == result_3

store.delete_document(result["_id"])

store.delete_documents([...])
```

## Issues

https://github.com/opea-project/GenAIComps/issues/1367

## Type of change

## Dependencies

`python-arango` is required for `ArangoDBStore`. Only checked upon instantiation.

## Tests

`test_arangodb.py` introduced in `tests/cores/storage`


aMahanna:
CC @rbrugaro

comps/cores/storages/arangodb.py
```diff
@@ -0,0 +1,288 @@
...
+            **kwargs: Additional arguments for saving the document.
+
+        Returns:
+            bool | dict: The result of the save operation.
+        """
+        try:
+            return self.collection.insert(doc, **kwargs)
+        except Exception as e:
+            logger.error(f"Failed to save document: {e}")
+            raise
```
aMahanna, line 98: Another question: Does OPEA prefer a catch & release implementation? Should we not re-raise but instead fail gracefully? Or should we completely remove the `try/except` clause and leave it up to the user to wrap?

aMahanna:
@rbrugaro 

> thanks @aMahanna! 2. prompt registry has search_by_keywork so maybe we can add that 3. 

Good catch. In favour of making this dynamic I've introduced `search` to `OpeaStore` (cc066d2c). Instead of restricting to just keyword search, I think we can make this dynamic enough via a `search_type` parameter. ArangoDB for example has both Vector Search and a Full-Text Search Engine built-in that I could eventually expand `ArangoDBStore` to include this functionality.

> Opea_prompt_microservice.py, opea_feedback_microservice.py… because they are specific to the data to be stored we had the definition and creation of data in same file like class FeedbackData , class ChatFeedback, def create_feeback_data, class PromptID, def PromptCreate… with the current abstraction of document where will the user put these definitions?

This one is tricky, because we do not want to restrict `OpeaStore` to _only_ handle these class types. As expected, if we start introducing Pydantic Models to this Client API, then it means we are restricting functionality to _only_ those objects. 

My personal opinion is that we should leave this for the user to handle in their own implementation, such as adding a validating step prior to invoking `OpeaStore` functions.

Happy to further discuss. 
 
> Also, I noticed in the redis integration (that was temporarily added in comp/agent/src/integrations/storage and will be refactored once this is done) there are sync and async redis_client with different API. I believe not all data storage providers have that but if needed should these functions in OpeaStore have the async counterparts?

Yep good point! Added this in `OpeaStore` only for now, and will introduce this in `ArangoDBStore` soon: 142ae235

rbrugaro:
thanks @aMahanna! I like the flexibility of search_type. In the case of custom it can be use for any metadata filtering right? 
I am thinking in hybrid search where user may want metadata filter and then vector search, would user use this custom search method and then call the retriever similarly search function? at the moment the similarity search are part of the retriever containers not core/common.., so wondering if best approach is to keep this opea store mostly KV Store focus (not vector search) or should we expand here (now or future) too? 
in a fiinance use case we are working on we use both jointly, we use KV for full docs and then embed summaries of doc chunks for content similarity retrieval..

Totally agree on not restricting opea store to a small set of class types for flexibility. But is there value on keeping around some of the very common class definitions for reusability? basically not constrain the OpeaStore but still make available the data class definitions if user wants to reuse prompt class, feedback class, or history class...

aMahanna:
@rbrugaro 

> thanks @aMahanna! I like the flexibility of search_type. In the case of custom it can be use for any metadata filtering right? 

Correct! It would be up to the `OpeaStore` subclass to implement a solution for handling `search_type=="custom"`. Most likely using `kwargs` to some extent.

> I am thinking in hybrid search where user may want metadata filter and then vector search, would user use this custom search method and then call the retriever similarly search function?  at the moment the similarity search are part of the retriever containers not core/common.., so wondering if best approach is to keep this opea store mostly KV Store focus (not vector search) or should we expand here (now or future) too?

For something as dynamic as this, I would have to think about and get back to you. We could start with OpeaStore as a KV-focused implementation, and wait for community feedback if enhancements make sense.

> Totally agree on not restricting opea store to a small set of class types for flexibility. But is there value on keeping around some of the very common class definitions for reusability? basically not constrain the OpeaStore but still make available the data class definitions if user wants to reuse prompt class, feedback class, or history class...

Ah understood! Yes that makes sense. Added here: 54f9fba4

mkbhanda:
@aMahanna please address DCO issue https://github.com/apps/dco

comps/cores/common/storage.py
```diff
@@ -0,0 +1,257 @@
+# Copyright (C) 2024 Intel Corporation
```
ftian1, line 1: not sure if the copyright is correct. @rbrugaro could you pls check? it should be the submitter's company copyright or intel copyright?

ftian1:
> @chensuyue I see the pre-commit.ci has added the intel copyright header to the arangoDB integrations that @aMahanna is contributing. does that need to be replaced?

just saw your comments. if it's auto added, it should be wrong.

comps/cores/storages/arangodb.py
ftian1: I would suggest to leave the upper level or the caller to deal with it.

ftian1:
for the documentation, it should be from developer view on explaining the interface and how to add new db. 

there is not good place to put it yet. you can put it in a RFC in opea-projects/docs repo, or here with sth like developer.md. Later on, we can make corresponding changes based on this.

chensuyue:
> @chensuyue I see the pre-commit.ci has added the intel copyright header to the arangoDB integrations that @aMahanna is contributing. does that need to be replaced?

If the new file has no default license hander, CI will automatically add it. So need PR submitter to update it according the real company.

aMahanna:
@ftian1 

> for the documentation, it should be from developer view on explaining the interface and how to add new db.
> 
> there is not good place to put it yet. you can put it in a RFC in opea-projects/docs repo, or here with sth like developer.md. Later on, we can make corresponding changes based on this.

Thanks. `developer.md` file added here: 1ae83e8a

comps/cores/common/storage.py
aMahanna: I've left this copyright untouched because it's completely independent of ArangoDB. Only modified the copyright for 1 file: https://github.com/opea-project/GenAIComps/commit/b52e811239a771f56a55f5e2b850432e181fcfff

comps/cores/storages/arangodb.py
aMahanna: Understood, @rbrugaro do you have any preference on this?

aMahanna:
Unrelated question: I've noticed this error before https://github.com/opea-project/GenAIComps/actions/runs/14247987788/job/39933707012?pr=1493#step:8:1490

We're dealing with the same issue with our Retriever service. Could we get more info on why this is happening?

comps/cores/storages/arangodb.py
rbrugaro: I think the catch and re-raise you have for called to handle is OK. I think the original microservices for chat history, feedback, prompt also dealt with it this way

comps/cores/common/storage.py
rbrugaro: @aMahanna even if this class is not ArangoDB specific since you are contributing you can put your copyright and get credit for it :) OPEA is an open source project so anyone can contribute to common features or vendor integrations.

rbrugaro:
@aMahanna the jaeger issue is related to telemetry. Please have a look at test failure: https://github.com/opea-project/GenAIComps/actions/runs/14247987788/job/39935693223?pr=1493#step:8:19
storages/test_arangodb.py::test_crud ERROR
storages/test_arangodb.py::test_initialize_connection FAILED

aMahanna:
> @aMahanna the jaeger issue is related to telemetry. Please have a look at test failure: https://github.com/opea-project/GenAIComps/actions/runs/14247987788/job/39935693223?pr=1493#step:8:19 storages/test_arangodb.py::test_crud ERROR storages/test_arangodb.py::test_initialize_connection FAILED

Thank you @rbrugaro! I see now, `no module named arango`. It seems to not recognize `python-arango` as a dependency, which makes sense. Are we allowed to add test dependencies? I see a `requirements.txt` at the top level of the repo. Should we introduce some kind of `test-requirements.txt` file?

rbrugaro:
@aMahanna why separate test-requirements.txt. is this dependency unique to tests? i see python-arango in arangodb.py

rbrugaro:
@aMahanna Understood. That's a good point. I think we can add to `requirements.txt `for now so we can run tests and if @chensuyue @feng-intel have preference to add a `test-requirements.txt` then we can move.

aMahanna:
@rbrugaro sounds good! added: 6db1eb6e

ftian1 (approved):
the overall looks good to me. but pls fix the CI failure like DCO issue.

comps/cores/common/storage.py
```diff
@@ -0,0 +1,239 @@
...
+    """The OpeaStore class serves as the base class for all Storage APIs.
+    It provides a unified interface and foundational attributes that every derived Storage API inherits and extends.
+
+    Attributes:
+        name (str): The name of the component (e.g 'arangodb', 'redis', 'mongodb', etc.)
+        description (str): A brief description of the component's functionality.
+        config (dict): A dictionary containing configuration parameters for the component.
+    """
+
+    def __init__(self, name: str, description: str = "", config: dict = {}):
```
letonghan, line 22: Suggest to add the `health_check` function in class `OpeaStore`, and implement it in `arangodb.py`. So that it will be easier for components like feedback_management to do health check when initializing services.
The health check should be a simple check to validate that the connection to database works well.

letonghan (commented):
Hi @aMahanna , thanks for your contribution!
one suggestion here, please consider to refine it!

lkk12014402 (approved):
LGTM

requirements.txt
```diff
@@ -13,6 +13,7 @@ opentelemetry-sdk
 Pillow
 prometheus-fastapi-instrumentator
 pypdf
+python-arango
```
lvliang-intel, line 16: @chensuyue is this package in the BOM list?

lvliang-intel (approved):

requirements.txt
rbrugaro: @aMahanna are there other dependency packages you need for dataprep/retriever PRs? please give a heads up so we can start adding those to BOM list while we get the PRs completed.

aMahanna: Good point @rbrugaro:

- For `ArangoDBStore`, only `python-arango` is required
- For ArangoDB Dataprep & Retriever services, `langchain-arangodb` is required
    - FYI here are the dependencies of `langchain-arangodb`: https://github.com/arangoml/langchain-arangodb/blob/main/libs/arangodb/pyproject.toml#L14-L20

rbrugaro:
@aMahanna I just merged main, if tests pass please resolve the comments to unblock the PR and we can get this merged. seems already for 3 approvals.

comps/cores/common/storage.py
aMahanna: Thank you, added: 07ac5319

aMahanna:
rebased via instructions: https://github.com/opea-project/GenAIComps/pull/1493/checks?check_run_id=40138654962

Squash & Merge should be ready after CI.

Thank you all

Issue #1465, opened 2025-03-26
opea/comps-base:latest image become too large


chensuyue:
opea/comps-base:latest image become much larger than before, from ~150M to ~3G. Need to investigate the rationality.
https://hub.docker.com/layers/opea/comps-base/latest/images/sha256-1c16dd64bee82731f6e8bc730c8e21d6766f82233d1aa1d23ca3c14e8b231cbc 

![Image](https://github.com/user-attachments/assets/03cac76a-d041-42ad-8137-8c0051d49469)

chensuyue:
![Image](https://github.com/user-attachments/assets/4249659c-e086-42ff-85a8-37476f7632c3)

xiguiw:
@eero-t 

I know you are working on https://github.com/opea-project/GenAIExamples/pull/1369
Any ideas about this?

ZePan110:
Remove langchain-huggingface from requirement by PR  https://github.com/opea-project/GenAIComps/pull/1505

eero-t:
@xiguiw IMHO this warrants filing upstream bug about `langchain-huggingface` increasing image size 20x (because according to @chensuyue's screenshot, it pulls in Nvidia drivers, which is also license issue, not just size one).

chensuyue:
We have root cause this issue, we should remove `langchain-huggingface` from the main requirements.

PR #1444, opened 2025-03-24
Fix Deprecation warnings in logs service logs

ZePan110:
## Description

Fix Deprecation warnings in logs service logs

The `HuggingFaceBgeEmbeddings` class deprecated in langchain and there's warning like below:
```bash
/home/user/comps/dataprep/src/integrations/redis.py:210: LangChainDeprecationWarning: The class `HuggingFaceBgeEmbeddings` was deprecated in LangChain 0.2.2 and will be removed in 1.0. 
An updated version of the class exists in the :class:`~langchain-huggingface package and should be used instead. 
To use it run `pip install -U :class:`~langchain-huggingface` and import as `from :class:`~langchain_huggingface import HuggingFaceEmbeddings``.
```
This PR will update all of the `HuggingFaceBgeEmbeddings` class into `HuggingFaceEmbeddings`.

## Issues

https://github.com/opea-project/GenAIComps/issues/1434

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [x] Others (enhancement, documentation, validation, etc.)

## Dependencies

List the newly introduced 3rd party dependency if exists.

## Tests

Describe the tests that you ran to verify your changes.

PR #1505, opened 2025-04-01
Remove langchain-huggingface from requirement.

ZePan110:
## Description

Remove langchain-huggingface from requirement.

## Issues

## Type of change

## Dependencies

## Tests

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,6 @@ httpx
 kubernetes
 langchain
 langchain-community
-langchain-huggingface
 opentelemetry-api
 opentelemetry-exporter-otlp
 opentelemetry-sdk


eero-t (changes requested):
Please list issue #1465 in PR description, and tell in commit description reason for the removal ("to avoid 150MB => 3GB image size increase").

requirements.txt
```diff
@@ -7,7 +7,6 @@ httpx
 kubernetes
 langchain
 langchain-community
-langchain-huggingface
```
eero-t, line 10: Looking at commit introducing this (https://github.com/opea-project/GenAIComps/commit/3e559df866972631e41a495b0bea068217224f86), Import / use of this module was added to `dataprep` and `retrievers` components, so you would need to add that module to their requirements files instead, otherwise this would break building of them.

eero-t (changes requested):

requirements.txt
chensuyue: It's already there.

eero-t: Ah, sorry, those seems to be already using their own requirements, please set this as resolved:
```
$ git grep requirements.txt dataprep/ retrievers/ | grep -v md:
dataprep/src/Dockerfile:    pip install --no-cache-dir ${PIP_EXTRA_INDEX_URL} -r /home/user/comps/dataprep/src/requirements.txt && \
retrievers/src/Dockerfile:    pip install --no-cache-dir ${PIP_EXTRA_INDEX_URL} -r /home/user/comps/retrievers/src/requirements.txt && \
```

(Which makes me wonder why https://github.com/opea-project/GenAIComps/pull/1444 added that dep to here in the first place....)

eero-t (approved):

chensuyue (approved):

PR #1547, opened 2025-04-10
Remove Transformers versions from requirements.txt file

ashahba:
## Description
There are multiple `requirements.txt` files in the repo with pinned versions of `transformers` and the suggested versions known to have issues.

## Issues
N/A

## Type of change
Pip dependency version relaxation

- [X] Others (enhancement, documentation, validation, etc.)

## Dependencies
None

## Tests
There are CI tests for this.


comps/third_parties/video-llama/src/requirements.txt
```diff
@@ -29,7 +29,7 @@ timm
 torch==1.13.1 --index-url https://download.pytorch.org/whl/cpu
 torchaudio==0.13.1 --index-url https://download.pytorch.org/whl/cpu
 torchvision==0.14.1 --index-url https://download.pytorch.org/whl/cpu
-transformers==4.47.1
```
chensuyue, line 32: How about freeze to 4.48.0? Because torch is fixed, I am afraid the deps will become incompatible with each other when the transformers version continue upgraded.

ashahba: No need to freeze the version here, because `pip install` would handle dependency resolution and it's pretty mature these days.
The goal of this PR is to resolve Dependabot flag, but full upgrade to Transformers v4.48.0 or better is only guaranteed after Torch ecosystem is also upgraded to newer versions.
I have a PR to address that here #1551 
Once that PR is merged Transformers version will be even newer than 4.48.0.

chensuyue: `pip install` will handle dependency resolution when there are transformer version limit inside torch installation config, otherwise I don't think it will.

ashahba: Ok, let me check on that.

chensuyue (approved):

ZePan110 (approved):

lvliang-intel (approved):

PR #1622, opened 2025-04-24
Reduce runtime memory for multilang tts

Spycsh:
## Description

Reduce runtime memory for multilang tts
Remove the .git/ and histories from the downloaded HF model repo

## Issues

na

## Type of change

## Dependencies

na
## Tests

ut


eero-t (approved):
Approved.

PS. AFAIK `*` pattern should skip dot files & dot dirs,  so removing `.git` dir before `mv` is probably not needed.  And because original repo dir is removed afterwards, `.git` dir history should also go away with it, so I think `--depth` arg addition actually affects only the Git repo download size, not final disk usage.  It's still optimization, and I like this PR making extra data exclusion more explicit.

lvliang-intel (approved):

lvliang-intel (approved):

CICD-at-OPEA:
This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 7 days.

chensuyue (approved):

PR #1626, opened 2025-04-27
Delete NV related package in xtune

jilongW:
## Description

Delete NV related package when build images

## Issues

n/a

## Type of change

## Dependencies

xtune

## Tests


lvliang-intel (approved):

chensuyue (approved):

eero-t (commented):
IMHO better than removing the packages, is failing the build if there are such dependencies. Nvidia builds of components like PyTorch are likely to fail on init after their Nvidia deps are gone.

As to where such detection should be done, doing it once for all images (e.g. in CI) sounds more attractive, to avoid whack-a-mole games...

jilongW:
Xtune will install pytorch and intel_extension_for_pytorch. 
It is safe to remove nvidia components in xtune image.
But for other image, this method will cause error.

PR #1627, opened 2025-04-28
Architecture specific args for a few containers

ashahba:
## Description

This PR adds optional `build-arg`'s for the following Dockerfiles:
```
- comps/animation/src/Dockerfile
- comps/llms/src/faq-generation/Dockerfile
- comps/third_parties/bridgetower/src/Dockerfile
- comps/third_parties/gpt-sovits/src/Dockerfile
```

## Issues

Current Dockerfiles always defaults to `gpu` Torch installation while `CPU` users might not necessarily need extra packages. This had the advantage of building smaller containers for CPU users and faster builds.

For users interested to do `gpu` builds, all they need to do is build this image this way for example:
```
docker build --build-arg ARCH='gpu' -f comps/animation/src/Dockerfile. -t opea/faq-generation:latest
```

## Type of change

- extra `build-arg` option `ARCH` to build a few of our containers for either CPU or GPU where applicable.

## Dependencies

None

### DIFF
--- a/comps/animation/src/Dockerfile
+++ b/comps/animation/src/Dockerfile
@@ -14,7 +14,11 @@ ARG ARCH=cpu
 COPY comps /home/user/comps
 
 RUN pip install --no-cache-dir --upgrade pip setuptools && \
-    pip install --no-cache-dir -r /home/user/comps/animation/src/requirements.txt ;
+    if [ ${ARCH} = "cpu" ]; then \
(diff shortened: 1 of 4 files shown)


chensuyue (approved):

ZePan110 (approved):

eero-t:
Well, Torch default is not even `gpu` in general, it's just Nvidia proprietary driver. So I think later on the `else` side would need to be diffentiated further (nvidia/intel/amd).  Preferably with some communication towards the Torch upstream, to get some guarantees that they keep different driver / license versions separate.

PR #1644, opened 2025-05-02
Agent: add support for remote server

alexsin368:
## Description

Add support for usage of a remote server inference endpoint. Supports Enterprise Inference.

Related PR: https://github.com/opea-project/GenAIExamples/pull/1900

## Issues

N/A

## Type of change

- [X] New feature (non-breaking change which adds new functionality)

## Dependencies

None

## Tests

Tested with AgentQnA consuming this agent microservice where chat completions were showing on the UI.


comps/agent/src/integrations/utils.py
```diff
@@ -66,7 +66,17 @@ def setup_chat_model(args):
             **params,
         )
     elif args.llm_engine == "openai":
-        llm = ChatOpenAI(model_name=args.model, request_timeout=args.timeout, **params)
+        if args.use_remote_service:
+            openai_endpoint = f"{args.llm_endpoint_url}/v1"
+            llm = ChatOpenAI(
+                openai_api_key=args.api_key,
```
louie-tsai, line 72: could we make the openai_api_key empty by default, so you don't need those if else in line 69 and 78. still passing openai_api_key and base but give empty string for none remote case.

comps/agent/src/integrations/utils.py
```diff
@@ -162,6 +172,10 @@ def get_args():
     parser.add_argument("--model", type=str, default="meta-llama/Meta-Llama-3-8B-Instruct")
     parser.add_argument("--llm_engine", type=str, default="tgi")
     parser.add_argument("--llm_endpoint_url", type=str, default="http://localhost:8080")
+    parser.add_argument("--api_key", type=str, default=None, help="API key to access remote server")
+    parser.add_argument(
+        "--use_remote_service", action="store_true", default=False, help="If using a remote server for LLM"
```
louie-tsai, line 177: we probably don't need this flag. if we have valid api_key, remote service should be on accordingly.

comps/agent/src/integrations/config.py
```diff
@@ -17,6 +17,12 @@
 if os.environ.get("llm_endpoint_url") is not None:
     env_config += ["--llm_endpoint_url", os.environ["llm_endpoint_url"]]
 
+if os.environ.get("api_key") is not None:
+    env_config += ["--api_key", os.environ["api_key"]]
+
+if os.environ.get("use_remote_service") is not None:
```
louie-tsai, line 23: probably don't need this env variable. just use api_key for remote service on/off judgement

alexsin368: removed use_remote_service

comps/agent/src/integrations/utils.py
alexsin368: removed use_remote_service

yinghu5 (approved):
LGTM, thank you!

comps/agent/src/integrations/utils.py
```diff
@@ -66,7 +66,17 @@ def setup_chat_model(args):
             **params,
         )
     elif args.llm_engine == "openai":
-        llm = ChatOpenAI(model_name=args.model, request_timeout=args.timeout, **params)
+        if args.api_key:
```
minmin-intel, line 69: Have you tested with OpenAI APIs?

alexsin368: Yes, I tested it using an OpenAI API. That's is what Enterprise Inference uses

minmin-intel: I was asking what about people who want to use GPT-4o from OpenAI? What base url should they use? Have you tested this scenario? Would also be good to add instructions in README about OpenAI vs. non-OpenAI remote LLM endpoints

alexsin368: For using GPT-4o from OpenAI, the AgentQnA example README has [instructions](https://github.com/opea-project/GenAIExamples/tree/main/AgentQnA#launch-on-xeon). The base URL is not needed in this case, the llm_endpoint_url env variable is set to http://${ip_address}:8080. I haven't tested this because I don't have an OpenAI API key and credits. If you share yours with me, I can try it out.

Added instructions in README on OpenAI models and models running from remote LLM endpoints

comps/agent/src/integrations/utils.py
louie-tsai: that could be easy solved with one line 
`openai_endpoint=args.llm_endpoint_url if args.llm_endpoint_url is None else args.llm_endpoint_url+ '/v1'`
also line 59-79 should be merged into one ChatOpenAI call

comps/agent/src/integrations/utils.py
```diff
@@ -66,7 +66,17 @@ def setup_chat_model(args):
             **params,
         )
     elif args.llm_engine == "openai":
-        llm = ChatOpenAI(model_name=args.model, request_timeout=args.timeout, **params)
+        if args.use_remote_service:
+            openai_endpoint = f"{args.llm_endpoint_url}/v1"
```
louie-tsai, line 70: that could be easy solved with one line
openai_endpoint=args.llm_endpoint_url if args.llm_endpoint_url is None else args.llm_endpoint_url+ '/v1'
also line 59-79 should be merged into one ChatOpenAI call

alexsin368: Do you mean regardless of args.llm_engine, we just do the following? By the way llm_endpoint_url is default to "http://localhost:8080". I think the new code would look like this:

`
openai_endpoint=None if args.llm_endpoint_url is "http://localhost:8080" else args.llm_endpoint_url+ '/v1'
llm = ChatOpenAI(
                openai_api_key=args.api_key,
                openai_api_base=openai_endpoint,
                model_name=args.model,
                request_timeout=args.timeout,
                **params,
            )
`

minmin-intel (approved):

comps/agent/src/integrations/utils.py
louie-tsai: looks good. that way, we don't need to have 3 call paths to call ChatOpenAI.  the main differences are values of input argument.

alexsin368: had to add back logic for llm_engine being set to "vllm" or "tgi", otherwise the hpu test fails. please review new implementation. test is passing now

alexsin368:
need help resolving DCO. steps i followed are not working

comps/agent/src/integrations/utils.py
louie-tsai: thanks. it looks much simple.

louie-tsai (approved):
looks good to me

alexsin368:
ready to merge!

Issue #1695, opened 2025-05-14
[Bug] can NOT ingest pdf with table_strategy=fast to dataprep on redis


lianhao:
### Priority

P3-Medium

### OS type

Ubuntu

### Hardware type

Xeon-GNR

### Installation method

- [x] Pull docker images from hub.docker.com
- [ ] Build docker images from source
- [ ] Other
- [ ] N/A

### Deploy method

- [ ] Docker
- [x] Docker Compose
- [ ] Kubernetes Helm Charts
- [ ] Other
- [ ] N/A

### Running nodes

Single Node

### What's the version?

docker image digest: 755e137d3feae1010c9db25a27e680c0ea754b755e22401a3e077a892c6e79a4

### Description

Can NOT ingest pdf files with table_strategy=fast, running the following command failed with internal server error http status code:

curl http://localhost:6007/v1/dataprep/ingest      -X POST     -H "Content-Type: multipart/form-data"     -F "files=@./tests/ingest_dataprep.pdf" -F "process_table=true" -F "table_strategy=fast"

### Reproduce steps

1. Start the opea/dataprep service on redis DB
2. curl http://localhost:6007/v1/dataprep/ingest      -X POST     -H "Content-Type: multipart/form-data"     -F "files=@./tests/ingest_dataprep.pdf" -F "process_table=true" -F "table_strategy=fast"

### Raw log

```shell
[2025-05-14 04:15:14,039] [    INFO] - opea_dataprep_microservice - [ ingest ] Base mode                                                                                  [197/1999]
[2025-05-14 04:15:14,040] [    INFO] - redis_dataprep - [ redis ingest] File ingest_dataprep.pdf does not exist.
[2025-05-14 04:15:15,563] [   ERROR] - opea_dataprep_microservice - Error during dataprep ingest invocation: can only concatenate list (not "NoneType") to list
INFO:     127.0.0.1:54254 - "POST /v1/dataprep/ingest HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1696, opened 2025-05-14
dataprep: return same type of data in get_tables_result

lianhao:
## Description

Return the same type of data in get_tables_result() to fix issue #1695

## Issues

Fixes #1695.

## Type of change

List the type of change like below. Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Dependencies

List the newly introduced 3rd party dependency if exists.

## Tests

Describe the tests that you ran to verify your changes.

PR #1712, opened 2025-05-20
Fix dataprep ingest bug with table_strategy=fast

letonghan:
## Description

Fix dataprep ingest bug with table_strategy=fast
add content check for table_chunks in all related db.py

## Issues

Fixes https://github.com/opea-project/GenAIComps/issues/1695

## Type of change

## Dependencies

None

## Tests

Local tested


lianhao:
Should https://github.com/opea-project/GenAIComps/pull/1696 be a more quick way to fix this?

letonghan:
> Should #1696 be a more quick way to fix this?

I think both of these two PRs are needed.
#1696 fixes the return value of `get_tables_result`, make sure the function returns the right value.
#1712 fixes the exception process in integrations, so dataprep will not crash even if it gets a null value.

lvliang-intel (approved):

Issue #1725, opened 2025-05-21
[Bug] Microservice docsum-vllm-gaudi test hang on with "stream":false


chensuyue:
### Priority

P1-Stopper

### OS type

Ubuntu

### Hardware type

Gaudi2

### Installation method

- [x] Pull docker images from hub.docker.com
- [ ] Build docker images from source
- [ ] Other
- [ ] N/A

### Deploy method

- [x] Docker
- [x] Docker Compose
- [ ] Kubernetes Helm Charts
- [ ] Other
- [ ] N/A

### Running nodes

Single Node

### What's the version?

main branch 

### Description

Microservice docsum-vllm-gaudi test hang on with "stream":false  
https://github.com/opea-project/GenAIComps/actions/runs/15152729725/job/42604747277?pr=1714#step:5:4425 

### Reproduce steps

cd tests 
bash llms/test_llms_doc-summarization_vllm_on_intel_hpu.sh  

### Raw log

```shell
[ docsum-vllm-gaudi ] HTTP status is 200. Checking content...
+ '[' 200 -eq 200 ']'
+ echo '[ docsum-vllm-gaudi ] HTTP status is 200. Checking content...'
++ curl -s -X POST -d '{"messages":"2024年9月26日，北京——今日，英特尔正式发布英特尔® 至强® 6性能核处理器（代号Granite Rapids），为AI、数据分析、科学计算等计算密集型业务提供卓越性能。", "max_tokens":32, "language":"zh", "stream":false}' -H 'Content-Type: application/json' http://192.168.122.251:10506/v1/docsum
++ tee /home/sdp/opea-actions-runner/_work/GenAIComps/GenAIComps/tests/docsum-vllm-gaudi.log
Error: Process completed with exit code 124.
```

PR #1790, opened 2025-06-11
Daily update vLLM&vLLM-fork version

XuehaoSun:
## Description

Daily update vLLM&vLLM-fork version

## Issues

## Type of change

## Dependencies

## Tests


yinghu5:
link vllm CPU update https://github.com/opea-project/GenAIExamples/pull/2061

.github/env/_build_image.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+export VLLM_VER=v0.6.1
```
eero-t, line 5: Current Dockerfiles use `v0.8.3` for the CPU version, only OpenVINO Dockerfile uses the ancient `v0.6.1` vLLM.

=> Either add 3rd variable for OpenVINO version, or use same version for it as for CPU.

.github/workflows/_comps-workflow.yml
```diff
@@ -75,15 +75,12 @@ jobs:
...
-              git clone https://github.com/vllm-project/vllm.git vllm-openvino
-              cd ./vllm-openvino && git checkout v0.6.1 && git rev-parse HEAD && cd ../
+              git clone -b ${VLLM_VER} --single-branch https://github.com/vllm-project/vllm.git
           fi
           if [[ $(grep -c "vllm-gaudi:" ${docker_compose_yml}) != 0 ]]; then
-              git clone https://github.com/HabanaAI/vllm-fork.git && cd vllm-fork
-              VLLM_VER=v0.6.6.post1+Gaudi-1.20.0
-              echo "Check out vLLM tag ${VLLM_VER}"
-              git checkout ${VLLM_VER} &> /dev/null && cd ../
+              git clone -b ${VLLM_FORK_VER} --single-branch https://github.com/HabanaAI/vllm-fork.git
```
eero-t, line 83: Unless Git history is needed, use `--depth 1` to same bandwidth, disk space & time.

Same applies to all other updated Git clone operations in this PR.

tests/agent/build_vllm_gaudi.sh
```diff
@@ -5,13 +5,8 @@ function build_vllm_docker_images() {
...
     cd $WORKDIR
     echo $WORKPATH
-    if [ ! -d "./vllm" ]; then
-        git clone https://github.com/HabanaAI/vllm-fork.git
-    fi
-    cd ./vllm-fork
-    VLLM_VER=v0.6.6.post1+Gaudi-1.20.0
-    echo "Check out vLLM tag ${VLLM_VER}"
-    git checkout ${VLLM_VER} &> /dev/null
+    VLLM_FORK_VER=v0.6.6.post1+Gaudi-1.20.0
```
eero-t, line 8: Why these and the other scripts are not including this variable from the shell script added for defining them?

.github/env/_build_image.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
eero-t, line 1: For current purpose, e.g. `_vllm_versions.sh` seems more appropriate name.

(Exporting variables does not need to be slowed down by using interactive Bash shell instead of just `/bin/sh`.)

tests/lvms/test_lvms_vllm_on_intel_hpu.sh
```diff
@@ -18,9 +18,8 @@ function build_docker_images() {
     cd $WORKPATH
     echo $(pwd)
 
-    git clone https://github.com/HabanaAI/vllm-fork.git
-    cd ./vllm-fork/
-    git checkout f78aeb9da0712561163eddd353e3b6097cd69bac # revert this to habana_main when https://github.com/HabanaAI/vllm-fork/issues/1015 is fixed
+    VLLM_FORK_VER=f78aeb9da0712561163eddd353e3b6097cd69bac # revert this to habana_main when https://github.com/HabanaAI/vllm-fork/issues/1015 is fixed
+    git clone -b ${VLLM_FORK_VER} --single-branch https://github.com/HabanaAI/vllm-fork.git && cd ./vllm-fork
```
eero-t, line 22: The linked issue was marked as fixed already ~3 months ago, and you cannot use `-b` option with individual commits, only with branch and tag names.  Please fix.

eero-t (changes requested):

tests/lvms/test_lvms_vllm_on_intel_hpu.sh
XuehaoSun: fix in [f12616c](https://github.com/opea-project/GenAIComps/pull/1790/commits/f12616ca3b503a52a3522d9042317c6d88839c78)

.github/env/_build_image.sh
XuehaoSun: fix in [f12616c](https://github.com/opea-project/GenAIComps/pull/1790/commits/f12616ca3b503a52a3522d9042317c6d88839c78)

comps/third_parties/vllm/src/build_docker_vllm.sh
```diff
@@ -35,19 +35,14 @@ fi
 
 # Build the docker image for vLLM based on the hardware mode
 if [ "$hw_mode" = "hpu" ]; then
-    git clone https://github.com/HabanaAI/vllm-fork.git && cd vllm-fork
-    VLLM_VER=v0.6.6.post1+Gaudi-1.20.0
-    echo "Check out vLLM tag ${VLLM_VER}"
-    git checkout ${VLLM_VER} &> /dev/null
+    source $(git rev-parse --show-toplevel)/.github/env/_vllm_versions.sh
```
eero-t, line 38: This is nice as all scripts can use the same line regardless of where the script is (easy to grep etc), and it continues working even when the script is moved.

However, this does not work when user gets sources through some other means that Git, e.g. by downloading them from the releases page: https://github.com/opea-project/GenAIComps/releases

I don't think supporting building these images without Git checkout to be important enough use-case, to change this, but it's possible that somebody will file a bug about it.

If you're worried about that, you need to replace git invocations with hard-coded path. :-/

.github/env/_build_image.sh
eero-t: It would be good to update OpenVINO support, first to something closer to v0.8.1, and eventually to newer vLLM version using the new vLLM OpenVINO plugin project: https://github.com/vllm-project/vllm-openvino

But that's definitely something out of scope for this PR.

chensuyue:
llms_doc-summarization_vllm_on_intel_hpu
llms_faq-generation_vllm
Those 2 tests failed with known issue, test got stuck. https://github.com/opea-project/GenAIComps/issues/1725

PR #1808, opened 2025-06-20
fix: ArangoDB Retriever `ARANGO_SEARCH_MODE`

aMahanna:
## Description

Fixing the use of an ArangoDB Retriever environment variable: `ARANGO_SEARCH_TYPE` is renamed to `ARANGO_SEARCH_MODE`.

This is because the `RetrieverRequest` object in `api_protocol.py` already has `search_type` as a parameter, which conflicts with a specific `search_type` parameter of the `ArangoVector` class used in the ArangoDB Retriever Service.

## Type of change


rbrugaro:
Good catch @aMahanna . Now that you are at it, can you also add a reference to Arango README in the main retriever [README.md](https://github.com/aMahanna/GenAIComps/blob/fix-retriever-search-type/comps/retrievers/README.md) ? I thnk the main dataprep README.md is also missing it

